PR #4, opened 2018-03-22
ios binding and demo

kenarsa:


.gitignore
```diff
@@ -1 +1,4 @@
 *.pyc
```
amiranhari: Please have a .gitignore in the demo/ios and put the related iOS entries there. Also maybe we should take a look at existing .gitignore of iOS projects like this:
https://gist.github.com/macbaszii/4475403

.gitignore
```diff
@@ -1 +1,4 @@
...
+demo/ios/PorcupineDemo.xcodeproj/xcuserdata/
```
amiranhari: Please have a .gitignore in the demo/ios and put the related iOS entries there. Also maybe we should take a look at existing .gitignore of iOS projects like this:
https://gist.github.com/macbaszii/4475403

binding/ios/PorcupineManager.swift
```diff
@@ -0,0 +1,131 @@
+//
```
amiranhari: Please fo block comments use the /* */ as described here: https://developer.apple.com/library/content/documentation/Xcode/Reference/xcode_markup_formatting_ref/ComentBlock.html

binding/ios/PorcupineManager.swift
```diff
@@ -0,0 +1,131 @@
...
+    private var porcupine: OpaquePointer?
+
+    private let audioCallback: AudioQueueInputCallback = {
+        userData, queue, bufferRef, startTimeRef, numPackets, packetDescriptions in
+
+        let porcupineManager: PorcupineManager = Unmanaged<PorcupineManager>.fromOpaque(userData!).takeUnretainedValue()
+        let pcm = bufferRef.pointee.mAudioData.assumingMemoryBound(to: Int16.self)
+        var result = false
+
+        pv_porcupine_process(porcupineManager.porcupine, pcm, &result)
```
amiranhari: Should we check the status returned by pv_porcupine_process and throw accordingly?

binding/ios/PorcupineManager.swift
```diff
@@ -0,0 +1,131 @@
...
+
+        AudioQueueStart(queue, nil)
+
+        isListening = true
+    }
+
+    /// Stops recording and releases resources acquired by Porcupine.
+    func stop() {
+        if !isListening { return }
+
```
amiranhari: Maybe I'm wrong, but I think we need to drecycle some of the resources we are getting from AudioQueue; otherwise, we might have memory leaks!

demo/ios/PorcupineDemo/AppDelegate.swift
```diff
@@ -0,0 +1,51 @@
...
+//
+
+import UIKit
+
+@UIApplicationMain
+class AppDelegate: UIResponder, UIApplicationDelegate {
+
+    var window: UIWindow?
+
+
```
amiranhari: Do we really need this empty functions? If we are using them maybe we can just delete them! Is there any specific reason we need to keep them around?

binding/ios/PorcupineManager.swift
kenarsa: I think that's for markup. xcode inserts // and looking at Foundation classes seems like that is the preferred method.

binding/ios/PorcupineManager.swift
```diff
@@ -0,0 +1,131 @@
...
+            throw PorcupineManagerError.OutOfMemoryError
+        case PV_STATUS_INVALID_ARGUMENT:
+            throw PorcupineManagerError.InvalidArgumentError
+        default:
+            return
+        }
+    }
+
+    
+    /// Initializer.
```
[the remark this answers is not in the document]
kenarsa: please see above

.gitignore
kenarsa: agreed

demo/ios/PorcupineDemo/ViewController.swift
```diff
@@ -0,0 +1,105 @@
...
+                keywordFilePath: keywordFilePath!,
+                sensitivity: 0.5,
+                keywordCallback: keywordCallback)
+
+            do {
+                try porcupineManager.start()
+            } catch let ex {
+                let alert = UIAlertController(
+                    title: "Alert",
+                    message: "Failed to initialized Porcupine " + ex.localizedDescription,
```
amiranhari: I'm not sure but just show something that is user-friendly if it is user-facing. Maybe something like "Something went wrong..."

kenarsa: Agreed

demo/ios/PorcupineDemo/ViewController.swift
```diff
@@ -84,7 +84,7 @@ class ViewController: UIViewController, UIPickerViewDelegate, UIPickerViewDataSo
             } catch let ex {
                 let alert = UIAlertController(
                     title: "Alert",
-                    message: "Failed to initialized Porcupine " + ex.localizedDescription,
+                    message: "Something went wrong" + ex.localizedDescription,
```
amiranhari: PLease remove the exception's message here.

demo/ios/.gitignore
```diff
@@ -0,0 +1,19 @@
...
+  
+# Exclude user-specific XCode 3 and 4 files
+*.mode1
+*.mode1v3
+*.mode2v3
+*.perspective
+*.perspectivev3
+*.pbxuser
+*.xcworkspace
+xcuserdata
```
amiranhari: Add a new line.

PR #7, opened 2018-04-11
Multi key

kenarsa:


binding/python/porcupine.py
```diff
@@ -39,15 +39,28 @@ class PicovoiceStatuses(Enum):
     class CPorcupine(Structure):
         pass
 
-    def __init__(self, library_path, model_file_path, keyword_file_path, sensitivity):
+    def __init__(
+            self,
+            library_path,
+            model_file_path,
```
amiranhari: you can keep the interface the same and in the body of your function have
`if isinstance(model_file_path, list):
     model_file_paths = model_file_path
  else:
    model_file_paths = list(model_file_path)
`

Same for sensitivity.

kenarsa: that's an option. I prefer it this way as the argument names go with semantics.

demo/python/porcupine_demo.py
```diff
@@ -185,13 +192,13 @@ def _default_library_path():
...
-        if not args.keyword_file_path:
-            raise ValueError('keyword file path is missing')
+        if not args.keyword_file_paths:
+            raise ValueError('keyword file paths are missing')
 
         PorcupineDemo(
             library_path=args.library_path,
             model_file_path=args.model_file_path,
-            keyword_file_path=args.keyword_file_path,
+            keyword_file_paths=args.keyword_file_paths.split(','),
```
amiranhari: do this so to also strip the whitespaces: `[x.strip() for x in my_string.split(',')]`

binding/android/.idea/misc.xml
```diff
@@ -5,11 +5,12 @@
     <option name="myDefaultNotNull" value="android.support.annotation.NonNull" />
```
amiranhari: I don't think we should push this file to repo. Are we ignoring the whole .idea directory in .gitignore?

demo/python/porcupine_demo.py
kenarsa: good point. addressed.

binding/android/.idea/misc.xml
kenarsa: No. only these two sub-folders:

/.idea/workspace.xml
/.idea/libraries

why?

binding/ios/PorcupineManager.swift
```diff
@@ -74,9 +81,26 @@ class PorcupineManager {
     ///   - keywordCallback: Callback to be executed after wake word detection.
     init(modelFilePath: String, keywordFilePath: String, sensitivity: Float, keywordCallback: @escaping (() -> Void)) {
```
amiranhari: Please check the parameters. I think we have duplicate modelFilePath

binding/ios/PorcupineManager.swift
```diff
@@ -74,9 +81,26 @@ class PorcupineManager {
...
+
+    
+    /// Initializer for multiple keyword use case.
+    ///
+    /// - Parameters:
+    ///   - modelFilePath: Absolute path to file containing model parameters.
+    ///   - keywordFilePaths: Absolute paths to keyword files.
+    ///   - sensitivities: List of sensitivity parameters for each keyword.
+    ///   - keywordCallback: Callback to be executed after wake word detection.
+    init(modelFilePath: String, keywordFilePaths: [String], sensitivities: [Float], keywordCallback: @escaping ((Int32) -> Void)) {
```
amiranhari: Also here we don't need modelFilePath only modelFilePaths.

PR #29, opened 2018-06-01
Add extern "C" so the library can be linked with C++ compiler and files

veeableful:
I had a problem where I cannot link the `libpv_porcupine.so` library with C++ code and/or compiler. It would output something like this:
```
/tmp/ccDJaGzT.o: In function `main':
demo.c:(.text+0x6c): undefined reference to `pv_porcupine_init(char const*, char const*, float, pv_porcupine_object**)'
demo.c:(.text+0xee): undefined reference to `pv_porcupine_frame_length()'
demo.c:(.text+0x12d): undefined reference to `pv_porcupine_process(pv_porcupine_object*, short const*, bool*)'
demo.c:(.text+0x132): undefined reference to `pv_porcupine_frame_length()'
demo.c:(.text+0x181): undefined reference to `pv_porcupine_delete(pv_porcupine_object*)'
```
However, I find that if I add the `extern "C"` to the header file, the error goes away. I don't really know exactly why but I thought I'd share the fix here :stuck_out_tongue:

### DIFF
--- a/include/picovoice.h
+++ b/include/picovoice.h
@@ -17,6 +17,11 @@
 #ifndef PICOVOICE_H
 #define PICOVOICE_H
 
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
 /**
  * Audio sample rate accepted by Picovoice.
  */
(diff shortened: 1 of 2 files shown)


veeableful:
Hi @kenarsa, certainly! I've made the changes that you requested.

kenarsa:
Hi @veeableful! Thank you. Looks great.

PR #30, opened 2018-06-07
Added Python requirements

jhoelzl:
to run python demo/python/porcupine_demo.py

### DIFF
--- a/demo/python/requirements
+++ b/demo/python/requirements
@@ -1,0 +1,2 @@
+numpy
+pyaudio
--- a/demo/python/requirements
+++ b/demo/python/requirements
@@ -1,2 +1,0 @@
-numpy
-pyaudio
--- a/demo/python/requirements.txt
+++ b/demo/python/requirements.txt
@@ -1,0 +1,2 @@
+numpy
+pyaudio
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 pysoundfile==0.9.0
 enum34==1.1.6
-numpy==1.14.4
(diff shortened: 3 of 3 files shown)


kenarsa:
Thanks for PR. Unfortunately, CI fails when trying to install PyAudio. We had the same issue and ended up installing it using apt-get. How did you install PyAudio? pip?

jhoelzl:
Hi, i think you need to install portaudio before installing pyaudio with "pip".

kenarsa:
Got it. That's why we decided not to add it as a requirement and just use the aptitude package as also mentioned in official page here: https://people.csail.mit.edu/hubert/pyaudio/

jhoelzl:
Okay. But at least you could add "numpy" to requirements.txt.

kenarsa:
Definitely, let's do that, please :)
Sorry, I first commented and then updated it.

kenarsa:
@jhoelzl, could you please do your change like this? https://github.com/Picovoice/Porcupine/pull/34
Then, we don't need to worry about CI failing. Thanks. You can keep both dependencies.

jhoelzl:
@kenarsa i removed the version numbers. Or did I misunderstand you?

kenarsa:
Hi @jhoelzl 

That helps. But also I meant let's have a separate requirements file for demo under demo/python. Similar to what @MatthewScholefield  has done https://github.com/Picovoice/Porcupine/pull/34/files

It does not break CI anymore this way. Thanks a lot.

jhoelzl:
@kenarsa , okay i got it (finally) :)

Issue #27, opened 2018-05-30
python demo crashed


fquirin:
### Expected behavior
The python demo keeps running until I quit it.
### Actual behavior
The demo crashed with following error:
```
Traceback (most recent call last):
  File "demo/python/porcupine_demo.py", line 207, in <module>
    input_device_index=args.input_audio_device_index).run()
  File "demo/python/porcupine_demo.py", line 104, in run
    pcm = audio_stream.read(porcupine.frame_length)
  File "/usr/lib/python2.7/dist-packages/pyaudio.py", line 608, in read
    return pa.read_stream(self._stream, num_frames, exception_on_overflow)
IOError: [Errno -9981] Input overflowed

```
### Steps to reproduce the behavior
Start the demo on a Rasperry PI Zero with ReSpeaker 2-Mic-HAT with following settings:

`python demo/python/porcupine_demo.py --keyword_file_paths ./resources/keyword_files/grasshopper_raspberrypi.ppn --library_path ./lib/raspberry-pi/arm11/libpv_porcupine.so --input_audio_device_index 2
`

OS is Raspbian Stretch Lite.

kenarsa:
Could you do me a favor and tell me what is the CPU usage without running the demo and when running the demo? I think what is happening is that CPU usage goes up and audio buffer overflows. Let's verify it first.

fquirin:
top says 1% idle 55% when Picovoice is running. I'll try to keep an eye on that.

kenarsa:
OK. How long does it run before crashing? Is it intermittent?

fquirin:
Unfortunately the crash message has no timestamp so I have to check this again, but it looks like its happening after around 20min. Btw I have music running in the background (for testing and entertainment purposes) so this might play a role for the audio buffer maybe....
... uups it just crashed after 8min with CPU load at 55% (at least a few seconds before).

kenarsa:
Got it. We had this issue before on RPi zero. Could you please follow this issue: https://github.com/Picovoice/Porcupine/issues/10

I think you find the solution there.

fquirin:
I'm not sure if I got everything right in the other issue but I've changed the python_demo code to this (starting from line 82):

```
num_keywords = len(self._keyword_file_paths)
		
        def _audio_callback(in_data, frame_count, time_info, status):
			if frame_count >= porcupine.frame_length:
				pcm = struct.unpack_from("h" * porcupine.frame_length, in_data)
[...]

PR #32, opened 2018-06-13
added non-blocking version of python demo

fquirin:
see:
https://github.com/Picovoice/Porcupine/issues/27

Starting from the input-buffer overflow issue I've encountered on my raspberry Pi zero I've changed the demo to run pyaudio in non-blocking mode (basically it does a callback to the execution function instead of waiting for processing to finish). It also gives you the chance to add some custom actions to the callback which always crashed the program in the blocking version.

### DIFF
--- a/demo/python/porcupine_demo_non_blocking.py
+++ b/demo/python/porcupine_demo_non_blocking.py
@@ -70,7 +70,7 @@ class PorcupineDemo(Thread):
         self._model_file_path = model_file_path
         self._keyword_file_paths = keyword_file_paths
         self._sensitivity = float(sensitivity)
-        self.frame_length = frame_length
+        self._frame_length = frame_length
         self._input_device_index = input_device_index
 
         self._output_path = output_path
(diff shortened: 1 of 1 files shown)


kenarsa:
This is amazing. I think the community would love this as it is extremely useful for smaller devices such as RPi zero. I have few minor comments and only one major one. All can be addressed easily :)

Let read through my comments and address if necessary. Happy to discuss :)

fquirin:
Happy to address your comments, but ... I can't find them :sweat_smile: Am I missing something?

kenarsa:
Sorry, my mistake. Forgot to click on submit.

fquirin:
I've addressed all the comments and removed the frame_length parameter. I think you are right, it will not work anyway if it is not the same as inside porcupine. I put it in for testing, but as far as I remember changing the value resulted in no keyword detection all the time and then I just left it at the default value :see_no_evil: I wasn't entirely sure what happens inside porcupine though and thought it might be useful at some point. Guess not ^^.

kenarsa:
Thanks a lot :)

PR #34, opened 2018-06-13
Create requirements.txt for demo

MatthewScholefield:
Without numpy and pyaudio the demo won't work


MatthewScholefield:
Closing since duplicate of #30. Do consider adding pyaudio, however, inside the test folder like this to solve CI issues.

PR #35, opened 2018-06-13
Require keyword file argument in add_argument

MatthewScholefield:
This is cleaner and gives a better output


kenarsa:
Awesome! Thank you.

PR #57, opened 2018-07-24
Add Rust demo

veeableful:
I created Rust [bindings](https://crates.io/crates/porcupine-sys) for Porcupine and thought of sharing a demo of its usage here :grin:. Compiling and running it should be very simple if the user has already installed the native libraries in discoverable locations (e.g. `/usr/local/lib`).

There _is_ some problem with getting the desired microphone format under macOS though. I might have to convert the audio data (e.g. from float32) to Porcupine-compatible format before processing it.

### DIFF
--- a/demo/rust/Cargo.lock
+++ b/demo/rust/Cargo.lock
@@ -1,0 +1,541 @@
+[[package]]
+name = "aho-corasick"
+version = "0.6.6"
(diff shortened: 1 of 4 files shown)


kenarsa:
This is awesome. Give me some time to review the code (I am going to use this as an opportunity to learn  Rust!).

Could you please which platforms have you tested this demo on? Thanks in advance.

PR #58, opened 2018-07-28
Fix Rust demo on macOS

veeableful:
Update Rust demo to use default audio input and convert samples to Porcupine's format.
    
Previously, the demo was hard-coded to open microphone with Porcupine's format which fails on macOS. Now, the demo runs on both Linux and macOS.


kenarsa:
Thank you @veeableful  :)

PR #62, opened 2018-08-01
Outdated example for iOS.

fabioknoedtbcgdv:


README.md
```diff
@@ -299,6 +299,7 @@ var handle: OpaquePointer?
 
 let status = pv_porcupine_multiple_keywords_init(
     modelFilePath,
+    2, // Number of different keywords to monitor for, or keywordFilePaths.count
```
kenarsa: Let's just use keywordFilePaths.count if possible.

README.md
```diff
@@ -316,13 +317,13 @@ func getNextAudioFrame() -> UnsafeMutablePointer<Int16> {
 
 while true {
     let pcm = getNextAudioFrame()
-    var keyword_index: Int32 = -1
```
kenarsa: Do you think we can just replace "pv_porcupine_process" with "pv_porcupine_multiple_keywords_process"? Then it is coherent with how we initialized the object to do multi-keyword detection above. Thank you.

fabioknoedtbcgdv:
Made the changes you proposed.

kenarsa:
thank you @fabioknoedtbcgdv

PR #67, opened 2018-08-14
Update iOS Binding

herlihalim:
- Refactoring the code to follow Swift convention closer.
- Opening up class to have `public` access level, so it can be packaged in a framework easier in the future.
- Fixed keywords file paths for the demo project.


kenarsa:
This awesome. Thank you. Let me go through it and I'll get back to you in a couple of days. One thing is that we don't want to break people who are using the engine currently. If we have to do it next release (v1.5) would be a good time to merge breaking changes.

herlihalim:
Hi, I just noticed that the watchOS and iOS binding are different, I refactored them so both will use the same binding. The binary might be different, but I reckon the binding can be consistent.

Additionally, updated the demo as well, consolidating both iOS and watchOS under the same folder. I added a new `xcodeproj` file for the "no watchOS" version.

kenarsa:
Apologies for the delay again. I will make time to look into the changes this evening for sure. I have a quick question before getting started. Have you tested the changes on an iOS and watchOS device?

herlihalim:
Yes, that I did, the only reason I was making all the changes, including adding new errors, handling audio session, etc are because of the issues I found when trying to build a proof of concept app.

binding/ios/PorcupineManager.swift
```diff
@@ -11,150 +11,234 @@
...
+                return
+            }
+            if shouldBeListening {
+                try? audioEngine.start()
+            }
+        }
+    }
+}
+
+public struct WakeKeywordConfiguration {
```
kenarsa: If possible let's change it to either WakeWordConfiguration or KeywordConfiguration. I think these two are a bit more known.

binding/ios/PorcupineManager.swift
```diff
@@ -11,150 +11,234 @@
...
+    let name: String
+    let filePath: String
+    let sensitivity: Float
+
+    /// Initialiser for the wake word configuration.
```
kenarsa: Initialiser -> Initializer

binding/ios/PorcupineManager.swift
```diff
@@ -11,150 +11,234 @@
...
+
+        try audioSession.setCategory(AVAudioSessionCategoryRecord)
+        try audioSession.setMode(AVAudioSessionModeMeasurement)
+        try audioSession.setActive(true, with: .notifyOthersOnDeactivation)
+
+        let frameLength = UInt32(pv_porcupine_frame_length())
+        let sampleRate = Double(pv_sample_rate())
+
+        let recordingFormat = inputNode.inputFormat(forBus: busNumber)
+        let numberOfBuffer: UInt32 = 4
```
kenarsa: I think in watchOS binding I was reading 4 buffer at a time because of a technical issue with watchOS. Here is a brief explanation: Porcupine reads audio in 32 ms frames. watchOS for some reason does not work if you set the buffer size to anything below 100 ms. I can't recall why. So for watchOS, we had to put a hack to read  4 * 32 = 128 ms buffers. This is not ideal and if possible we should avoid it on iOS as it increases the delay. Let me know what you think.

binding/ios/PorcupineManager.swift
```diff
@@ -11,150 +11,234 @@
...
-        self.keywordCallback = nil
+    ///   - wakeKeywordConfigurations: Keyword configuration to use.
+    ///   - onDetection: Detection handler to call after wake word detection. The handler is executed on main thread.
+    /// - Throws: PorcupineManagerError
+    public convenience init(modelFilePath: String, wakeKeywordConfiguration: WakeKeywordConfiguration, onDetection: ((WakeKeywordConfiguration) -> Void)?) throws {
+        try self.init(modelFilePath: modelFilePath, wakeKeywordConfigurations: [wakeKeywordConfiguration], onDetection: onDetection)
+    }
+
+    deinit {
+        pv_porcupine_delete(porcupine)
```
kenarsa: I wonder if this should be done the other way around. First, stop listening and then delete the porcupine object. the reason is that there is a chance that the object is called after being deleted that can cause problems. I might be missing something here ...

binding/ios/PorcupineManager.swift
```diff
@@ -11,150 +11,234 @@
...
-    ///   - keywordFilePaths: Absolute paths to keyword files.
-    ///   - sensitivities: List of sensitivity parameters for each keyword.
-    ///   - keywordCallback: Callback to be executed after wake word detection.
-    init(modelFilePath: String, keywordFilePaths: [String], sensitivities: [Float], keywordCallback: @escaping ((Int32) -> Void)) {
-        self.modelFilePath = modelFilePath
-        self.keywordFilePaths = keywordFilePaths
-        self.sensitivities = sensitivities
-        self.multiKeywordCallback = keywordCallback
-        self.keywordCallback = nil
+    ///   - wakeKeywordConfigurations: Keyword configuration to use.
```
kenarsa: wakeKeywordConfigurations -> wakeKeywordConfiguration

herlihalim:
Cool, I'll have a look! Thanks. Will fix up things you mentioned.

Do you know where do you read about the watchOS buffer? It seems to work for me, will be great if I can find some reference to it. Worst come to worst, it's possible to just have different buffer size depending on the targetOS.

kenarsa:
Thanks a lot!

I am going to find out where. It might have been Apple's official documentation or StackOverflow. I will find the link ASAP. Quick question ... when you mentioned that it seems to work for you, did you mean the current code or refactored code? Yes, they will work for you :) But if you set the number of buffers to 1 it should not work on watchOS. or at least this is what I experienced.

binding/ios/PorcupineManager.swift
herlihalim: Yes, this work, the buffer size is set to approx 8196, it's bigger than previously assigned 5644 (0.128 * recordingFormat.sampleRate, which is 44100).

I have tested on both iOS and watchOS.

herlihalim:
All feedbacks are addressed. I have tested both iOS and watchOS using the same binding, i.e no modification at all. They're both working.

Thanks for the feedbacks!

binding/ios/PorcupineManager.swift
herlihalim: Oops, I misread this. I think I got what you mean now. I have to fix up the buffer size then.

kenarsa:
Looks great. Just had one comment about the buffer size. We need to get that right :) After that, it is all good and ready to be shipped. Thanks again for great work.

herlihalim:
Sorry for the delay, all fixed up now.

What I have done is abstracting out the "Input Engine". PorcupineManager still has the exact same interface, but now depending on the compilation target will automatically select different input engine.

For iOS/macOS it will use the AudioQueue input engine, so it can have the 32ms latency.
For watchOS it will use the AVAudioEngine input engine because that's the one available.

Please have a look. Cheers!

kenarsa:
Looks great. Thank you.

PR #82, opened 2018-09-20
Alsa Cpp Demo

ghost:
This is a demo using C++ and Alsa audio device with porcupine


demo/Alsa cpp/main.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+static char snd_device[] = "default";
+snd_pcm_t* capture_handle;
+snd_pcm_hw_params_t* hw_params;
+snd_pcm_info_t* s_info;
+unsigned int srate = 16000;
+unsigned int nchan = 1;
+
+
+int init_soundcard(char *devicename)
```
kenarsa: char * -> const char *

demo/Alsa cpp/compile.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
```
kenarsa: Thanks for making this cross-platform. Maybe we can use Cmake? You can actually write the whole thing in few lines of code then :)

demo/Alsa cpp/compile.sh
```diff
@@ -0,0 +1,46 @@
...
+then
+    SYSTEM="windows"
+    LIB="-L../../lib/windows/$MACHINE"
+fi
+
+if [ $SYSTEM = "Linux" ]
+then
+    if [[ $MACHINE == arm* ]];
+    then
+        SYSTEM="raspberry-pi"
```
kenarsa: This is a tricky one. Because this is the path for lib which is optimized for RPi zero. If someone wants to use RPi2/3 they need to use the other provided libs. Maybe leave this as an input argument. I am open to suggestions. Donno if there is an easy way to find the type of RPi reliably across models.

demo/Alsa cpp/compile.sh
```diff
@@ -0,0 +1,46 @@
...
+fi
+
+
+if [ $LIB != "NONE" ]
+then
+    g++ -o alsademo -I../../include $LIB main.cpp -lpv_porcupine -lasound
+    export LD_LIBRARY_PATH=../../lib/$SYSTEM/$MACHINE
+    echo "Build Complete"
+    echo "Use ./alsademo to begin"
+    echo "read the README.MD for any extra information"
```
kenarsa: README.md?

demo/Alsa cpp/compile.sh
```diff
@@ -0,0 +1,46 @@
...
+
+else
+    echo "Porcupine is not supported on $SYSTEM/$MACHINE yet!"
```
kenarsa: This is not essentially true. For example Android or other ARM Linux platforms ... Maybe we should allow the user to optionally enter the arguments. This is super easy with CMake.

demo/Alsa cpp/main.cpp
```diff
@@ -0,0 +1,216 @@
...
+ * Project: Porcupine Alsa
+ * Created Date: Tuesday, September 18th 2018, 7:41:55 pm
+ * Author: Greg
+ * -----
+ * Last Modified: Tue Sep 18 2018
+ * Modified By: Greg
+ * -----
+ * Copyright (c) 2018 Greg
+ * 
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
```
kenarsa: Could you please add the standard Apache 2 header? Since the repo is under Apache2 I am assuming you are OK releasing this code with same license?

demo/Alsa cpp/main.cpp
```diff
@@ -0,0 +1,216 @@
...
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN
+ * AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+ * WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ * 
+ */
+
+
+#include <iostream>
+#include <alsa/asoundlib.h>
+#include "../../include/pv_porcupine.h"
```
kenarsa: A cleaner way could be to add a -I in compile line with the path to include directory. But this would work as well.

demo/Alsa cpp/main.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+
+using namespace std;
+
+#define NO_ERROR	0
```
kenarsa: Maybe we can make an enum for this?

demo/Alsa cpp/main.cpp
```diff
@@ -0,0 +1,216 @@
...
+unsigned int srate = 16000;
+unsigned int nchan = 1;
+
+
+int init_soundcard(const char *devicename)
+{
+    //connect to and set parameters for the sound device
+    //returns any error setting or connecting
+
+	int err = 0;
```
kenarsa: extra space (I think!)

demo/Alsa cpp/main.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+    if (devicename)
+        strcpy(snd_device, devicename);
```
kenarsa: this is tricky! Because if the devicename is longer than memory allocated to snd_device you will overwrite memory! Do we need to copy ?

demo/Alsa cpp/compile.sh
oziee: downloading it now.. see how long it takes me to work this out :)

demo/Alsa cpp/main.cpp
oziee: think I will just make it "default" as a const.. that way the ./asoundrc or the /etc/asound.conf will be used

demo/Alsa cpp/main.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+	if ((err = snd_pcm_open(&capture_handle, snd_device,SND_PCM_STREAM_CAPTURE, 0)) < 0)
+	{
+		fprintf(stderr, "cannot open audio device %s (%s, %d)\n",snd_device, snd_strerror(err), err);
```
kenarsa: Maybe we can keep using std::cout to be consistent ?

demo/Alsa cpp/main.cpp
```diff
@@ -0,0 +1,216 @@
...
+        return START_PORCUPINE_ERROR;
+    }
+    return NO_ERROR;
+}
+
+int listen()
+{
+    //listen on the mic for the wake word
+
+	int err = 0;
```
kenarsa: extra space

demo/Alsa cpp/main.cpp
oziee: ??

demo/Alsa cpp/main.cpp
```diff
@@ -0,0 +1,216 @@
...
+    short wav_data[MAX_BUF_SIZE];
```
kenarsa: MAX_BUF_SIZE is really the buffer size. Maybe we can rename it to that? Also, we probably should grab this from pv_porcupine_frame_length() so if we change it at a later point to something like 256 the demo keeps working?

demo/Alsa cpp/main.cpp
kenarsa: fprintf is the C style logging. C++ developers use std::cout instead. I am OK with either. But like to use one consistently throughout the code.

demo/Alsa cpp/main.cpp
kenarsa: OK

demo/Alsa cpp/compile.sh
kenarsa: If too much headache, don't bother. But when you learn it you will use it forever!

kenarsa:
Looks really good. Had a few minor comments. Lets address and push it in. One thing which is cool to do is to compare the performance of this demo with Python one. For example on an RPi and see how much CPU this demo saves.

demo/Alsa cpp/compile.sh
```diff
@@ -0,0 +1,46 @@
...
+    else
+        SYSTEM="linux"
+        LIB="-L../../lib/linux/$MACHINE"
+    fi
+fi
+
+
+if [ $LIB != "NONE" ]
+then
+    g++ -o alsademo -I../../include $LIB main.cpp -lpv_porcupine -lasound
```
kenarsa: Add an optimization flag. Either -O2 or -O3 to make it even speedier!

PR #94, opened 2018-10-12
Fix typo in License

stephenwithav:
s/expect/except/.

"expect for" in court would likely not be legally binding.


kenarsa:
Goog catch :)

Thank you @stephenwithav

PR #103, opened 2018-10-22
implemented dotnet core 2.1 bindings

HeadhunterXamd:


kenarsa:
Looks great. I have a few questions/comments:

1- Is it possible to make a READMe.md for the binding? Something like this would do: https://github.com/Picovoice/Porcupine/tree/master/binding/python. I just want people to know what they need to install and how to run the unit test. 

2- Is WAV functionality part of some other library? If yes, we need to include the appropriate license information. If not, it will be available under Apache 2 which is great.

3- I wonder if it is possible to use this in a Xamarin project. Any insight? That can open the functionality to app world as well. Just a question for my own knowledge.

Thanks again for your contribution :)

HeadhunterXamd:
Sure I will add a README.md and look for the licensing (how to add it) for the WAV functionality (this was only added to use in the unit tests).
(I only added this because I didn't really 
I think it works in an xamarin project (this is written for general csharp not only dotnet core) only I haven't tested for that.

HeadhunterXamd:
The full source of the bindings are one csharp file, so it could be easy to just copy the code into an xamarin project to test it.

kenarsa:
Looks great. Merging!

PR #168, opened 2019-04-25
V1.6

kenarsa:


shaneboyar:
Looks like the new bindings are not backward-compatible old versions of the keyword files. You should make sure to reach out to paying clients with new keyword files so they can upgrade.

mehrdadfeller:
Commercial licenses are granted to customers for a specific version of the software. The licensing terms won't automatically grant rights to other versions of the software. To upgrade license rights to the latest software version, customers have to either pay an upgrade fee. However, if they have already purchased one of our annual technical support packages, they will receive upgrades at no additional charge during the support term.

PR #396, opened 2021-04-08
iOS SDK migrated to Cocoapods

laves:


binding/ios/Porcupine.swift
```diff
@@ -0,0 +1,180 @@
+//
+//  Copyright 2018-2021 Picovoice Inc.
```
kenarsa, line 2: just 2021

binding/ios/Porcupine-iOS.podspec
```diff
@@ -0,0 +1,27 @@
+Pod::Spec.new do |s|
```
kenarsa, line 1: point to the correct readme.md

binding/ios/Porcupine.swift
```diff
@@ -0,0 +1,180 @@
...
+//  specific language governing permissions and limitations under the License.
+//
+
+import PvPorcupine
+
+public enum PorcupineError: Error {
+    case invalidArgument(message:String)
+    case io
+    case outOfMemory
+    case recordingDenied
```
kenarsa, line 16: let's put these two in manager

demo/ios/Podfile
```diff
@@ -0,0 +1,7 @@
+source 'git@gitlab.com:picovoice/podspec-test.git'
```
kenarsa, line 1: remember remember

PR #422, opened 2021-05-20
Go SDK

laves:
Adding Go support for Porcupine and associated demos


binding/go/porcupine.go
```diff
@@ -0,0 +1,281 @@
...
+		if k == b {
+			return nil
+		}
+	}
+	return errors.New("Invalid built-in keyword.")
+}
+
+// Porcupine struct
+type Porcupine struct {
+	handle uintptr
```
dbartle, line 102: other ones have comments so this one should probably have one too, even if its kind of self-explanatory

binding/go/porcupine_test.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+
+		if result == 0 {
+			results = append(results, result)
+			t.Logf("Keyword triggered at %f", float64((i+1)*FrameLength)/float64(SampleRate))
+		}
+	}
+
+	if len(results) != 1 || results[0] != 0 {
+		t.Fatal("Failed to find keyword.")
```
dbartle, line 65: Would be useful to output the keyword name / file name here

binding/go/porcupine_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+func TestMultiple(t *testing.T) {
+
+	test_file, _ := filepath.Abs("../../resources/audio_samples/multiple_keywords.wav")
+
+	p := Porcupine{BuiltInKeywords: []BuiltInKeyword{
+		ALEXA, AMERICANO, BLUEBERRY, BUMBLEBEE,
+		GRAPEFRUIT, GRASSHOPPER, PICOVOICE, PORCUPINE,
+		TERMINATOR}}
+	expectedResults := []int{7, 0, 1, 2, 3, 4, 5, 6, 7, 8}
```
dbartle, line 82: Using the ints as the results isn't ideal, can we use the enum names instead. Maybe cast if necessary?

binding/go/porcupine_unix.go
```diff
@@ -0,0 +1,132 @@
+// Copyright 2021 Picovoice Inc.
```
dbartle, line 1: Not that important but "porcupine_posix" might be more apt than "unix"

binding/go/porcupine_unix.go
```diff
@@ -0,0 +1,132 @@
...
+	porcupine.handle = uintptr(ptrC[0])
+	return int(ret)
+}
+
+func (porcupine *Porcupine) nativeDelete() {
+	C.pv_porcupine_delete_wrapper(pv_porcupine_delete_ptr,
+		unsafe.Pointer(porcupine.handle))
+}
+
+func (porcupine *Porcupine) nativeProcess(pcm []int16) (int, int) {
```
dbartle, line 112: return PvStatus for 1st arg

dbartle: or return error if its not PV_SUCCESS?

binding/go/porcupine_windows.go
```diff
@@ -0,0 +1,91 @@
...
+		uintptr(unsafe.Pointer(&porcupine.handle)))
+
+	return int(ret)
+}
+
+func (porcupine *Porcupine) nativeDelete() {
+	delete_func.Call(porcupine.handle)
+}
+
+func (porcupine *Porcupine) nativeProcess(pcm []int16) (int, int) {
```
dbartle, line 68: PvStatus cast (or error) for status return variable

binding/go/porcupine.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+
+	if len(pcm) != FrameLength {
+		return -1, fmt.Errorf("Input data frame is wrong size")
+	}
+
+	// call process
+	ret, index := porcupine.nativeProcess(pcm)
+	if PvStatus(ret) != SUCCESS {
+		return -1, fmt.Errorf("Process audio frame failed with PvStatus: %d", ret)
```
dbartle, line 211: return the status instead of -1?

binding/go/porcupine.go
```diff
@@ -0,0 +1,281 @@
...
+// linearly-encoded. Porcupine operates on single-channel audio.
+// Returns a 0 based index if keyword was detected in frame. Returns -1 if no detection was made.
+func (porcupine *Porcupine) Process(pcm []int16) (int, error) {
+
+	if porcupine.handle == 0 {
+		return -1, fmt.Errorf("Porcupine has not been initialized or has been deleted.")
+	}
+
+	if len(pcm) != FrameLength {
+		return -1, fmt.Errorf("Input data frame is wrong size")
```
dbartle, line 205: put expected and actual frame lengths in error message

binding/go/porcupine.go
```diff
@@ -0,0 +1,281 @@
...
+	switch os := runtime.GOOS; os {
+	case "darwin":
+		return "mac"
+	case "linux":
+		return "linux"
+	case "windows":
+		return "windows"
+	default:
+		log.Fatalf("%s is not a supported OS", os)
+		return ""
```
dbartle, line 227: return error?

binding/go/porcupine.go
```diff
@@ -0,0 +1,281 @@
...
+	var libPath string
+	switch os := runtime.GOOS; os {
+	case "darwin":
+		libPath = fmt.Sprintf("embedded/lib/%s/x86_64/libpv_porcupine.dylib", osName)
+	case "linux":
+		libPath = fmt.Sprintf("embedded/lib/%s/x86_64/libpv_porcupine.so", osName)
+	case "windows":
+		libPath = fmt.Sprintf("embedded/lib/%s/amd64/libpv_porcupine.dll", osName)
+	default:
+		log.Fatalf("%s is not a supported OS", os)
```
dbartle, line 262: return error?

binding/go/porcupine.go
```diff
@@ -0,0 +1,281 @@
...
+
+// PvStatus type
+type PvStatus int
+
+// Possible status return codes from the Porcupine library
+const (
+	SUCCESS          PvStatus = 0
+	OUT_OF_MEMORY    PvStatus = 1
+	IO_ERROR         PvStatus = 2
+	INVALID_ARGUMENT PvStatus = 3
```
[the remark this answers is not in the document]
laves: doesn't porcupine never throw these states?

binding/go/porcupine.go
laves: first value is keyword index - I wouldn't want folks getting confused that they got a valid detection even if an error is returned. That's why I put the status code in the error.

binding/go/porcupine.go
laves: log fatal is equivalent to throwing a global error and then exiting the application. Since this is an internal function, the developer doesn't have access to the return values, so this is the best way to inform them that something has gone wrong.

laves: (ie the return doesn't actually get triggered, but the compiler needs it there)

binding/go/porcupine.go
laves: same as above

binding/go/porcupine.go
dbartle: makes sense

dbartle (approved):

binding/go/porcupine.go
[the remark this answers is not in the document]
dbartle: include to match header regardless

PR #433, opened 2021-06-08
Flutter auto-extract assets

laves:
- implemented auto extraction of given model and keyword Flutter assets
- tested on iOS and Android
- specified list of built-in keywords in copy script
- ran through dart format


binding/flutter/lib/porcupine.dart
```diff
@@ -88,33 +92,71 @@ class Porcupine {
...
+      if (keywordPaths[i] == null || keywordPaths[i] == "") {
+        throw new PvArgumentError(
+            "One of the provided keyword paths was empty.");
+      }
+
+      if (!await File(keywordPaths[i]!).exists()) {
+        try {
+          keywordPaths[i] = await _extractResource(keywordPaths[i]!);
+        } catch (_) {
+          throw new PvArgumentError(
```
dbartle, line 158: does dart have a concept of nested exceptions? there is value in not throwing away whatever the original exception was, even if we're translating it upstream to a "Pv***Error"

binding/flutter/lib/porcupine.dart
```diff
@@ -88,33 +92,71 @@ class Porcupine {
...
 
-    modelPath ??= _defaultModelPath;
+    if (modelPath == null || modelPath == "") {
+      modelPath = _defaultModelPath;
+    }
+
+    if (!await File(modelPath!).exists()) {
+      try {
+        modelPath = await _extractResource(modelPath);
+      } catch (_) {
```
dbartle, line 120: see nested exception comment below and add if possible

binding/flutter/lib/porcupine.dart
```diff
@@ -88,33 +92,71 @@ class Porcupine {
...
       }
 
       keywordPaths = List.filled(keywords.length, "");
       for (var i = 0; i < keywords.length; i++) {
         if (BUILT_IN_KEYWORDS.contains(keywords[i])) {
           keywordPaths[i] = _builtInKeywordPaths?[keywords[i]];
         } else {
-          throw new PvArgumentError("One or more keywords are not available by default. Available default keywords are: $BUILT_IN_KEYWORDS");
+          throw new PvArgumentError(
+              "One or more keywords are not available by default. Available default keywords are: $BUILT_IN_KEYWORDS");
```
dbartle, line 143: include the (first) specific keyword that failed in the error message

binding/flutter/lib/porcupine.dart
laves: Not that I can find. I can nest it as part of the message parameter of the PvArgumentError though. Mostly this call just serves to check that the file exists in the app bundle, which you can't do directly.

dbartle (approved):

PR #465, opened 2021-08-06
Add Rust wrapper and demos

ErisMik:
- Adds a Rust wrapper for the Porcupine library as well as both file and mic demos
- There are some unit tests for the wrapper. Manual testing was done on a a Dell XPS laptop for Linux + Windows, and a Macbook Air. Tests for ARM boards are in progress.

- You can verify the changes by installing Rust (https://rustup.rs/) and Clang (`libclang-dev` linux / https://github.com/llvm/llvm-project/releases/tag/llvmorg-12.0.1 windows), and running `cargo run --release -- <args>` in either of the demo dirs.

- Does not include any of the packaging logic, even the packaging of the core Porcupine library into the wrapper package.


binding/rust/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "pv_porcupine"
+version = "0.1.0"
```
laves, line 3: version 1.9.0 (the last digit is your package version, the first two are the version of Porcupine)

binding/rust/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "pv_porcupine"
```
laves, line 2: simply "porcupine" is good (most of our packages don't have the pv prefix)

binding/rust/src/porcupine.rs
```diff
@@ -0,0 +1,383 @@
...
+            let frame = frame.collect_vec();
+            if frame.len() == porcupine.frame_length() as usize {
+                let keyword_index = porcupine.process(&frame).unwrap();
+                if keyword_index >= 0 {
+                    results.push(keyword_index);
+                }
+            }
+        }
+
+        assert_eq!(results, [6, 0, 1, 2, 3, 4, 5, 6, 7]);
```
laves, line 381: It's better to have a list of keyword strings (e.g. ["porcupine", "alexa"...]). Check that list size against results, and then loop through results and check each result value against the input keyword list and the expected keyword list.  If the string gained from both don't match, then we fail.

binding/rust/src/util.rs
```diff
@@ -0,0 +1,167 @@
+/*
+    Copyright 2018-2021 Picovoice Inc.
```
laves, line 2: 2021

binding/rust/src/util.rs
```diff
@@ -0,0 +1,167 @@
...
+        let entry = entry.unwrap();
+        let path = entry.path();
+        let keyword = entry
+            .file_name()
+            .into_string()
+            .unwrap()
+            .split("_")
+            .next()
+            .unwrap()
+            .to_string();
```
laves, line 162: can this fail? if so, we should probably check if the entry conforms to our file name pattern and exists before adding it

demo/rust/filedemo/src/main.rs
```diff
@@ -0,0 +1,195 @@
+/*
+    Copyright 2018-2021 Picovoice Inc.
```
laves, line 2: 2021

demo/rust/filedemo/src/main.rs
```diff
@@ -0,0 +1,195 @@
...
+    let default_library_path = pv_library_path("../../../")
+        .into_os_string()
+        .into_string()
+        .unwrap();
+    let default_model_path = pv_model_path("../../../")
+        .into_os_string()
+        .into_string()
+        .unwrap();
+
+    let matches = App::new("Picovoice Porcupine Rust Mic Demo")
```
laves, line 71: mic demo -> file demo

demo/rust/filedemo/src/main.rs
```diff
@@ -0,0 +1,195 @@
...
+        .arg(
+            Arg::with_name("input_audio_path")
+            .long("input_audio_path")
+            .value_name("PATH")
+            .help("Absolute path to input audio file.")
```
laves, line 76: does it work with relatives?

demo/rust/filedemo/src/main.rs
```diff
@@ -0,0 +1,195 @@
...
+        )
+        .get_matches();
+
+    let input_audio_path = PathBuf::from(matches.value_of("input_audio_path").unwrap());
+    let library_path = PathBuf::from(matches.value_of("library_path").unwrap());
+    let model_path = PathBuf::from(matches.value_of("model_path").unwrap());
+
+    let keyword_paths: Vec<PathBuf> = {
+        if matches.is_present("keywords") {
+            let pv_keyword_paths = pv_keyword_paths(PathBuf::from("../../../"));
```
laves, line 135: should not have to fetch from relative, should be able to pass in built-in enum

demo/rust/micdemo/src/main.rs
```diff
@@ -0,0 +1,299 @@
+/*
+    Copyright 2018-2021 Picovoice Inc.
```
laves, line 2: 2021

demo/rust/micdemo/src/main.rs
```diff
@@ -0,0 +1,299 @@
...
+            .possible_values(&["Wasapi", "DSound", "WinMM", "CoreAudio", "SNDIO", "Audio4", "OSS", "PulseAudio", "Alsa", "Jack", "AAudio", "OpenSL", "WebAudio"])
+        )
+        .arg(
+            Arg::with_name("show_audio_devices")
+            .long("show_audio_devices")
+        )
+        .get_matches();
+
+    if matches.is_present("show_audio_devices") {
+        show_audio_devices();
```
laves, line 204: can you just return here instead of having a big else block?

demo/rust/micdemo/src/main.rs
```diff
@@ -0,0 +1,299 @@
...
+
+fn process_audio_chunk(
+    samples: &[i16],
+    porcupine: &mut Porcupine,
+    buffer: &mut Vec<i16>,
+    keywords: &[String],
+) {
+    buffer.extend_from_slice(samples);
+
+    while buffer.len() >= porcupine.frame_length() as usize && LISTENING.load(Ordering::SeqCst) {
```
laves, line 30: I think it's possible we could get a smaller buffer in a rare case. Best not to exit the recording loop, but just report that we got a small buffer, don't pass it to porcupine and continue to loop. If we get a larger buffer then frame_length, report it to console, pass as many subframes of it to porcupine as you can and then continue the loop.

demo/rust/micdemo/src/main.rs
```diff
@@ -0,0 +1,299 @@
...
+        let frame: Vec<i16> = buffer.drain(..porcupine.frame_length() as usize).collect();
+        let result = porcupine.process(&frame).unwrap();
+        if result >= 0 {
+            println!(
+                "[{}] Detected {}",
+                SystemTime::now()
+                    .duration_since(SystemTime::UNIX_EPOCH)
```
laves, line 37: Can you make the time more readable?

demo/rust/micdemo/src/main.rs
```diff
@@ -0,0 +1,299 @@
...
+
+use clap::{App, Arg, ArgGroup};
+use ctrlc;
+use miniaudio;
+use pv_porcupine::{pv_keyword_paths, pv_library_path, pv_model_path, Porcupine, KEYWORDS};
+use std::path::PathBuf;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::time::SystemTime;
+
+static LISTENING: AtomicBool = AtomicBool::new(true);
```
laves, line 20: a bit pedantic, but should probably set this to false first, then when you're ready to launch the recording loop, set to true, then set to false again

binding/rust/src/porcupine.rs
```diff
@@ -0,0 +1,383 @@
...
+                        ),
+                    ));
+                }
+            }
+
+            for sensitivity in sensitivities {
+                if !(0.0 <= *sensitivity && *sensitivity <= 1.0) {
+                    return Err(PorcupineError::new(
+                        PorcupineErrorStatus::ArgumentError,
+                        "A sensitivity value should be within [0, 1]",
```
laves, line 181: tell them which value was out of bounds (Sensitivity value of <> was invalid.)

binding/rust/src/porcupine.rs
```diff
@@ -0,0 +1,383 @@
...
+                        "Couldn't find model file at {}",
+                        model_path.as_ref().display()
+                    ),
+                ));
+            }
+
+            if keyword_paths.len() != sensitivities.len() {
+                return Err(PorcupineError::new(
+                    PorcupineErrorStatus::ArgumentError,
+                    "Number of keywords does not match the number of sensitivities",
```
laves, line 161: tell them how many keywords and how many sensitivities in the message

demo/rust/filedemo/src/main.rs
```diff
@@ -0,0 +1,195 @@
...
+
+    for frame in &audiosource.chunks(porcupine.frame_length() as usize) {
+        let frame = frame.collect_vec();
+        if frame.len() == porcupine.frame_length() as usize {
+            let keyword_index = porcupine.process(&frame).unwrap();
+            if keyword_index >= 0 {
+                println!(
+                    "[{}] Detected {}",
+                    SystemTime::now()
+                        .duration_since(SystemTime::UNIX_EPOCH)
```
laves, line 51: Can you make the time more readable?

laves (changes requested):

binding/rust/src/porcupine.rs
```diff
@@ -0,0 +1,383 @@
...
+
+impl PorcupineInner {
+    pub fn new<P: AsRef<Path>>(
+        library_path: P,
+        model_path: P,
+        keyword_paths: &[P],
+        sensitivities: &[f32],
+    ) -> Result<Self, PorcupineError> {
+        unsafe {
+            if !library_path.as_ref().exists() {
```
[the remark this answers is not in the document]
ErisMik: In this case `exists()` is actually a check to see if the file exists on disk. There's no optional parameters in Rust :)

binding/rust/src/porcupine.rs
```diff
@@ -0,0 +1,383 @@
...
+                return Err(PorcupineError::new(
+                    PorcupineErrorStatus::ArgumentError,
+                    &format!(
+                        "Couldn't find model file at {}",
+                        model_path.as_ref().display()
+                    ),
+                ));
+            }
+
+            if keyword_paths.len() != sensitivities.len() {
```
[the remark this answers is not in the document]
ErisMik: Nope, but in theory they could be empty, what's the behaviour we want there?

binding/rust/src/porcupine.rs
```diff
@@ -0,0 +1,383 @@
...
+            let pv_model_path = pathbuf_to_cstring(&model_path);
+            let pv_keyword_paths = keyword_paths
+                .iter()
+                .map(|keyword_path| pathbuf_to_cstring(keyword_path))
+                .collect::<Vec<_>>();
+            let pv_keyword_paths_ptrs = pv_keyword_paths
+                .iter()
+                .map(|keyword_path| keyword_path.as_ptr())
+                .collect::<Vec<_>>();
+            let mut cporcupine = std::ptr::null_mut();
```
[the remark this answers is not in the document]
ErisMik: They can fail. Right now they fail unrecoverable-y, by gracefully do you mean something code could handle at a later level?

binding/rust/src/porcupine.rs
```diff
@@ -0,0 +1,383 @@
...
+
+            let status = pv_porcupine_init(
+                pv_model_path.as_ptr(),
+                pv_keyword_paths.len() as i32,
+                pv_keyword_paths_ptrs.as_ptr(),
+                sensitivities.as_ptr(),
+                addr_of_mut!(cporcupine),
+            );
+            if status != PicovoiceStatuses::SUCCESS {
+                panic!(
```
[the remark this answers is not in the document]
ErisMik: Exceptions in Rust are either returning the `Result` type with `Err` so the caller handles, or hard failing with an `abort!`, `panic!`, `.unwrap()` or `.expect()`. I'll have this return ass an Err.

binding/rust/src/porcupine.rs
```diff
@@ -0,0 +1,383 @@
...
+                cporcupine,
+                lib,
+                sample_rate,
+                frame_length,
+            });
+        }
+    }
+
+    pub fn process(&self, pcm: &[i16]) -> Result<i32, PorcupineError> {
+        if pcm.len() as i32 != self.frame_length {
```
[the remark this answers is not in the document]
ErisMik: Nope :D 
Rust won't let you do that!

binding/rust/src/porcupine.rs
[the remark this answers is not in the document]
laves: Cool! Empty lists, we'd want to error out and let the user know they didn't provide keyword_paths and/or sensitivities. That could probably be handled on the individual list checks though and this check could be moved down to occur after those checks.

binding/rust/src/porcupine.rs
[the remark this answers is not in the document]
ErisMik: In this case the `&[..]` syntax is called a slice, which doesn't represent a type per-say but a "view" into a supported type (Like a vector). Since Rust operates entirely on moves/references, the "borrow-checker" can guarantee at compile-time that this view will never be `null` at runtime.

binding/rust/src/porcupine.rs
```diff
@@ -0,0 +1,383 @@
...
+                    "Found a frame length of {} Expected {}",
+                    pcm.len(),
+                    self.frame_length
+                ),
+            ));
+        }
+
+        let mut result: i32 = -1;
+        let status = unsafe {
+            let pv_porcupine_process: PvPorcupineProcessFn = self
```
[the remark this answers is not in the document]
ErisMik: It's possible, but I'd have to play with lifetimes and I'll admit I'm not an expert there. I'll give a go if you think that it is important.

binding/rust/src/porcupine.rs
```diff
@@ -0,0 +1,383 @@
...
+        }
+
+        return Ok(result);
+    }
+}
+
+unsafe impl Send for PorcupineInner {}
+unsafe impl Sync for PorcupineInner {}
+
+impl Drop for PorcupineInner {
```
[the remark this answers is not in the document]
ErisMik: Yes, it's guaranteed to get called and guaranteed to only get called once :)

binding/rust/src/porcupine.rs
```diff
@@ -0,0 +1,383 @@
...
+                .lib
+                .get(b"pv_porcupine_delete")
+                .expect("Failed to load function symbol from Porcupine library");
+            pv_porcupine_delete(self.cporcupine);
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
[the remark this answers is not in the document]
ErisMik: It's pretty common to ship the tests as part of the package in Rust. They don't have to be in the same file though, I'll move them out if you'd prefer.

binding/rust/src/porcupine.rs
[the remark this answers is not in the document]
laves: by graceful I just mean that we should provide a readable error/exception to them saying "could not find x in Porcupine library". That way if they've passed in some BS library or an older version it'll provide a better indication of why it failed. That said, maybe the error return is already fairly readable - I just wouldn't want a super hard crash.

binding/rust/src/porcupine.rs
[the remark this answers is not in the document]
laves: yes, I think returning on Err on any init fail would be good. Including the library functions failing to load as mentioned above

binding/rust/src/porcupine.rs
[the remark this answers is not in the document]
laves: yeah, definitely worth it for process function

binding/rust/src/porcupine.rs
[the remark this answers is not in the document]
laves: nice!

binding/rust/src/porcupine.rs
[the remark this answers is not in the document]
laves: let's move them to a separate file, but yeah, if it's standard to ship the tests along with, then we can do that as well.

demo/rust/micdemo/src/main.rs
ErisMik: With a package dependency, sure :)

binding/rust/src/porcupine.rs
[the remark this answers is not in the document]
ErisMik: It's similar to the `RAII` philosophy I think it's called in C++, but the Rust memory model makes it way easier.

demo/rust/micdemo/src/main.rs
laves: dang, really? Might be worth it for the demo. The timestamp just looks pretty gnarly. Either that or don't show the timestamp.

demo/rust/filedemo/src/main.rs
laves: For this actually, you may want to print when in the file it occurred, e.g. "Detected porcupine at 3:40", rather than the timestamp.

demo/rust/micdemo/src/main.rs
ErisMik: That is what the `buffer` is for, to handle different sized inputs. Unless I'm missing an edge case I don't see?

laves: ah okay, so buffer is always a multiple of porcupine frame length? if there is any possibility that may not be true we should handle it. If not, great!

binding/rust/src/util.rs
```diff
@@ -84,92 +88,97 @@ fn base_library_path() -> PathBuf {
...
-    path.push("../");
-    path.push("lib/common/porcupine_params.pv");
+    path.push(DEFAULT_RELATIVE_MODEL_PATH);
     return path;
 }
 
 #[cfg(target_os = "macos")]
-fn keyword_path_subdir() -> PathBuf {
-    return PathBuf::from("mac");
+fn keyword_path_subdir() -> String {
```
laves, line 125: even though this string does represent the keyword subdir, maybe we should just call the function pv_platform because sometimes it's a directory sometimes we use it in the keyword file name to okay the platform tag.

binding/rust/build.rs
```diff
@@ -0,0 +1,62 @@
+/*
+    Copyright 2021-2021 Picovoice Inc.
```
laves, line 2: oh sorry, I meant just put "Copyright 2021 Picovoice Inc.". When it's the one year, we don't need to have a dash.

laves (changes requested):

demo/rust/micdemo/src/main.rs
ErisMik: I think what you are thinking of the `buffer` is actually the `samples`. The `samples` is what is returned to us from the audio library callback. We add `samples` another buffer `buffer`, and then while `buffer` is longer than a single frame, we "drain" (popping iterator) the buffer `frame_length` items. `buffer` persists across callback calls, so whether the `samples` is larger or smaller than a single `frame_length` doesn't matter.

laves (approved):

binding/rust/README.md
```diff
@@ -0,0 +1,91 @@
...
+```
+In the above example, we've initialzed the engine to detect the built-in wake word "Porcupine". Built-in keywords are contained in the package with the `BuiltinKeywords` enum type.
+
+Porcupine can detect multiple keywords concurrently:
+```rust
+let porcupine: Porcupine = PorcupineBuilder::new_with_keywords(&[BuiltinKeywords::Porcupine, BuiltinKeywords::Blueberry, BuiltinKeywords::Bumblebee])
+    .init().expect("Unable to create Porcupine");
+```
+
+To detect non-default keywords, use `PorupineBuilder`'s `new_with_keyword_paths` method instead:
```
laves, line 52: *custom keywords

laves: Also, mention here that they should pass in a .ppn file just to be a bit more clear

binding/rust/README.md
```diff
@@ -0,0 +1,91 @@
...
+```rust
+let porcupine: Porcupine = PorcupineBuilder::new_with_keyword_paths(&["/absolute/path/to/keyword/one", "/absolute/path/to/keyword/two"])
+    .init().expect("Unable to create Porcupine");
+```
+
+The sensitivity of the engine can be tuned per keyword using the `sensitivities` method:
+```rust
+let porcupine: Porcupine = PorcupineBuilder::new_with_keywords(&[BuiltinKeywords::Porcupine, BuiltinKeywords::Bumblebee])
+    .sensitivities(&[0.2f32, 0.42f32])
```
laves, line 61: You may also want to show that you can set the model file as well since that's required for changing the language

binding/rust/copy.sh
```diff
@@ -0,0 +1,17 @@
+echo "Preppring dir ..."
```
laves, line 1: *Preparing

demo/rust/README.md
```diff
@@ -0,0 +1,116 @@
...
+NOTE: The working directory for the following `Cargo` commands is:
+
+```console
+porcupine/demo/rust/filedemo  # File Demo
+porcupine/demo/rust/micdemo  # Mic Demo
+```
+
+### File Demo
+
+The file demo uses Porcupine to scan for keywords in a wave file. The demo is mainly useful for quantitative performance benchmarking against a corpus of audio data.
```
laves, line 20: *wav file

laves (changes requested):

PR #466, opened 2021-08-06
iOS binding upgrade

ksyeo1010:
- Update iOS binding to use VoiceProcessor

### DIFF
--- a/binding/ios/PorcupineManager.swift
+++ b/binding/ios/PorcupineManager.swift
@@ -117,7 +117,7 @@ public class PorcupineManager {
         }
 
         // Only check if it's denied, permission will be automatically asked.
-        if try !voiceProcessor.hasPermissions() {
+        guard try voiceProcessor.hasPermissions() else {
             throw PorcupineManagerError.recordingDenied
         }
         
@@ -144,6 +144,10 @@ public class PorcupineManager {
     /// Callback to run after after voice processor processes frames.
     private func audioCallback(frameLength: UInt32, pcm: UnsafePointer<Int16> ) {
         guard self.porcupine != nil else {
+            return
+        }
+        
+        guard frameLength == Porcupine.frameLength else {
             return
         }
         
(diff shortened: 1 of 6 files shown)


binding/ios/PorcupineManager.swift
```diff
@@ -30,28 +30,7 @@ public class PorcupineManager {
...
-                if result >= 0 {
-                    DispatchQueue.main.async {
-                        self.onDetection?(result)
-                    }
-                }
-            } catch {
-                print("\(error)")
-            }
-        }
+        self.voiceProcessor = VoiceProcessor()
```
laves, line 33: our other voice processors all conform to the singleton pattern. I think you should try to do the same with ios-voice-processor. Reference the flutter-voice-processor for how that's done. If we do instances, then a user could potentially create several voice processors, which would not function properly.

binding/ios/PorcupineManager.swift
```diff
@@ -138,14 +117,15 @@ public class PorcupineManager {
...
             throw PorcupineManagerError.recordingDenied
         }
         
-        try audioSession.setCategory(AVAudioSession.Category.playAndRecord, options: [.mixWithOthers, .defaultToSpeaker, .allowBluetooth])
-        
-        try audioInputEngine.start()
+        try voiceProcessor.start(
+            frameLength: Porcupine.frameLength,
+            sampleRate: Porcupine.sampleRate,
+            audioCallback: self.audioCallback
```
laves, line 127: Is there a way you can subscribe/unsubscribe callbacks from the voice processor? That way, with the singleton pattern, you can just add audio callbacks on-the-fly and remove them when you no longer want to be notified.

laves (changes requested):

laves (changes requested):
Can you remove the process function from `Porcupine.Swift` that uses the UnsafePointer? I don't think that function is used anymore.

PR #470, opened 2021-08-12
Support go bindings for rpi, jetson and beaglebone

ksyeo1010:
- Added binding support rpi,jetson and beaglebone support for go bindings
- Removed warnings generated by static checks


binding/go/porcupine.go
```diff
@@ -246,17 +247,65 @@ func (porcupine *Porcupine) Process(pcm []int16) (keywordIndex int, err error) {
...
+		return "linux", "x86_64"
+	} else if runtime.GOARCH == "arm64" {
+		archInfo = "-aarch64"
+	}
+
+	cmd := exec.Command("cat", "/proc/cpuinfo")
+	cpuInfo, err := cmd.Output()
+
+	if err != nil {
+		log.Fatalf("Failed to get CPU details: %s, defaulting to linux x86_64", err.Error())
```
laves, line 278: log.Fatalf kills the application as soon as possible, but runs this print first. In this case, I believe we actually do want it to fail instead of reverting to linux, so maybe just print failed to get cpu details.

binding/go/porcupine.go
```diff
@@ -246,17 +247,65 @@ func (porcupine *Porcupine) Process(pcm []int16) (keywordIndex int, err error) {
...
+	case "0xd03":
+		return "raspberry-pi", "cortex-a53" + archInfo
+	case "0xd07":
+		return "jetson", "cortex-a57" + archInfo
+	case "0xd08":
+		return "raspberry-pi", "cortex-a72" + archInfo
+	case "0xc08":
+		return "beaglebone", ""
+	default:
+		log.Fatalf(
```
laves, line 305: In this case we don't want it to fail, so use log.Printf

laves (changes requested):

binding/go/porcupine.go
```diff
@@ -230,11 +231,11 @@ func (porcupine *Porcupine) Delete() error {
 func (porcupine *Porcupine) Process(pcm []int16) (keywordIndex int, err error) {
 
 	if porcupine.handle == 0 {
-		return -1, fmt.Errorf("Porcupine has not been initialized or has been deleted.")
+		return -1, fmt.Errorf("Porcupine has not been initialized or has been deleted")
 	}
 
 	if len(pcm) != FrameLength {
-		return -1, fmt.Errorf("Input data frame size (%d) does not match required size of %d", len(pcm), FrameLength)
+		return -1, fmt.Errorf("input data frame size (%d) does not match required size of %d", len(pcm), FrameLength)
```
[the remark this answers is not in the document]
ksyeo1010: Go expected errors to be in lower case since the usually follow other context. See: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

laves (approved):

PR #474, opened 2021-08-12
Rust Github Action

ErisMik:


.github/workflows/rust.yml
```diff
@@ -0,0 +1,47 @@
+name: Rust
+
+on:
+  push:
+    branches: [ master ]
+    paths:
+      - 'binding/rust/**'
+      - 'lib/linux/**'
+      - 'lib/mac/**'
+      - 'lib/raspberry-pi/**'
```
laves, line 10: should add jetson and beaglebone too right? our .NET SDK doesn't have jetson and beaglebone, so those weren't included

laves (changes requested):

laves (changes requested):
Can we include `keyword_files` for supported platforms and `lib/common/porcupine_params.pv` too?

laves (approved):

PR #475, opened 2021-08-13
Java binding support for arm

ksyeo1010:
- Added Java bing support for RPi 2, 3, 4, Jetson and BeagleBone
  - Added JNI
  - Fixed central repository
- Updated Readme


binding/java/src/ai/picovoice/porcupine/Utils.java
```diff
@@ -34,6 +34,8 @@
     private static final String ENVIRONMENT_NAME;
     private final static Logger logger = Logger.getLogger(Logger.GLOBAL_LOGGER_NAME);
 
+    private static String linuxArch;
```
laves, line 37: To keep it more flexible and clean, why don't we have a static function in the static block below that gets OS and Arch and then we'll use those in the rest of the functions.

laves (changes requested):

binding/java/src/ai/picovoice/porcupine/Utils.java
```diff
@@ -121,24 +121,78 @@ private static Path extractResources(Path jarPath) throws IOException {
...
+                    return "cortex-a53" + archInfo;
+                case "0xd07":
+                    return "cortex-a57" + archInfo;
+                case "0xd08":
+                    return "cortex-a72" + archInfo;
+                case "0xc08":
+                    return "";
+                default:
+                    throw new RuntimeException(
+                            String.format("Platform architecture with CPU Part (%s) is not supported by Porcupine.", cpuPart)
```
laves, line 179: don't we usually revert to some generic arm build at this point instead of throwing a fatal?

laves (changes requested):

laves (approved):

PR #488, opened 2021-09-07
updated rust demo to use pvrecorder

ksyeo1010:
- updated rust demo to use pvrecorder

### DIFF
--- a/demo/rust/micdemo/src/main.rs
+++ b/demo/rust/micdemo/src/main.rs
@@ -62,10 +62,7 @@ fn porcupine_demo(
     println!("Listening for wake words...");
 
     let mut audio_data = Vec::new();
-    // Spin loop until we receive the ctrlc handler
     while LISTENING.load(Ordering::SeqCst) {
-        std::hint::spin_loop();
-        std::thread::sleep(std::time::Duration::from_millis(10));
         let mut pcm = vec![0; recorder.frame_length()];
         recorder
             .read(&mut pcm)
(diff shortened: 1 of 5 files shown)


demo/rust/micdemo/src/main.rs
```diff
@@ -72,79 +46,67 @@ fn porcupine_demo(
...
-
     ctrlc::set_handler(|| {
         LISTENING.store(false, Ordering::SeqCst);
     })
     .expect("Unable to setup signal handler");
 
+    println!("Listening for wake words...");
+
+    let mut audio_data = Vec::new();
     // Spin loop until we receive the ctrlc handler
```
ErisMik, line 65: This comment does not apply anymore

demo/rust/micdemo/src/main.rs
```diff
@@ -72,79 +46,67 @@ fn porcupine_demo(
...
     while LISTENING.load(Ordering::SeqCst) {
         std::hint::spin_loop();
```
ErisMik, line 67: No longer needed (Not a spin loop anymore)

demo/rust/micdemo/src/main.rs
```diff
@@ -72,79 +46,67 @@ fn porcupine_demo(
...
         std::thread::sleep(std::time::Duration::from_millis(10));
```
ErisMik, line 68: Also not needed anymore, since the `pv_recorder` read blocks

demo/rust/micdemo/src/main.rs
```diff
@@ -72,79 +46,67 @@ fn porcupine_demo(
...
+    println!("\nStopping...");
+    recorder.stop().expect("Failed to stop audio recording");
+
+    if !output_path.is_none() {
+        let wavspec = hound::WavSpec {
+            channels: 1,
+            sample_rate: porcupine.sample_rate(),
+            bits_per_sample: 16,
+            sample_format: hound::SampleFormat::Int,
+        };
```
ErisMik, line 97: We can remove the `if` statement and move this block into the `Some(..)` path of the `match` statement. Otherwise we are checking for "someness" twice.

ErisMik: Actually I think a more idiomatic way would be to drop the `match` statement entirely since we don't do anything with the the `none` branch, you can use the `if let` syntax for this:

```rust
if let Some(output_path) = output_path {
...
}
```
This is `true` if output_path is `Some` and then gives us the inner value of the output_path `Option` inside the block, and `false` otherwise.

ErisMik (changes requested):
Some pedantic stuff and some minor cleanup but looks good!

ErisMik (approved):

PR #515, opened 2021-11-03
V2.0 react binding

mrrostam:
updated the react binding and demo to use the new web binding

### DIFF
--- a/binding/react/src/usePorcupine.ts
+++ b/binding/react/src/usePorcupine.ts
@@ -73,23 +73,6 @@ export function usePorcupine(
     return false;
   };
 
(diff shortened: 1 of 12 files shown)


binding/react/src/usePorcupine.ts
```diff
@@ -17,7 +17,6 @@ import {
   PorcupineHookArgs,
   PorcupineWorker,
   PorcupineWorkerFactory,
-  PorcupineWorkerResponse,
 } from './porcupine_types';
 
 export function usePorcupine(
```
ksyeo1010, line 22: line 43, porcupineWorker is not used, do we don't really need it?

ksyeo1010 (approved):

PR #522, opened 2021-11-05
GitHub action for the react demo

mrrostam:


.github/workflows/react_demos.yml
```diff
@@ -0,0 +1,56 @@
+name: React Demos
+
+on:
+  push:
```
ErisMik, line 4: This lets us manually trigger the actions from GitHub and is a huge help when testing new actions
```suggestion
  workflow_dispatch:
  push:
```

ErisMik (approved):

.github/workflows/react_demos.yml
mrrostam: Good idea, i would create a dummy pull request for such a thing!

PR #526, opened 2021-11-09
V2.0 flutter

ksyeo1010:
- updated flutter binding and demo
- moved away from flutter-ffi to native Java/Swift modules

To test on iOS: copy binding/ios/Porcupine-iOS.podspec to root dir.


binding/flutter/android/src/main/java/ai/picovoice/flutter/porcupine/PorcupinePlugin.java
```diff
@@ -11,23 +11,185 @@
...
+    channel.setMethodCallHandler(this);
+  }
+
+  @Override
+  public void onMethodCall(@NonNull MethodCall call, @NonNull Result result) {
+    Method method;
+    try {
+      method = Method.valueOf(call.method.toUpperCase());
+    } catch (IllegalArgumentException e) {
+      method = Method.DEFAULT;
```
laves, line 58: You should probably just throw here instead of having a .DEFAULT enum

binding/flutter/lib/porcupine.dart
```diff
@@ -92,121 +106,116 @@ class Porcupine {
...
-      } catch (_) {
-        throw new PvArgumentError(
-            "Could not find model file at path '$modelPath'. If this is a packaged asset, ensure you have added it to your pubspec.yaml file.");
-      }
-    }
-
-    if (keywords != null && keywordPaths != null) {
-      throw new PvArgumentError(
-          "Both 'keywords' and 'keywordPaths' were set. Only one of the two arguments may be set for initialization.");
+      } catch (e) {/* don't do anything on fail */}
```
laves, line 138: hmmm...not sure about this. I know we're trying to allow auto-extraction in Flutter and native, but there must be a more explicit way of checking this so that it doesn't just do a blind catch.

binding/flutter/lib/porcupine.dart
```diff
@@ -92,121 +106,116 @@ class Porcupine {
...
       List<double>? sensitivities}) async {
-    if (!_resourcesExtracted) {
-      await _extractPorcupineResources();
-      _resourcesExtracted = true;
+    if (accessKey == null || accessKey == "") {
+      throw PorcupineInvalidArgumentException("'accessKey' must be set.");
     }
 
     if (modelPath == null || modelPath == "") {
       modelPath = _defaultModelPath;
```
laves, line 132: we can probably let empty model path through and let native fill in default model, right?

binding/flutter/lib/porcupine.dart
```diff
@@ -92,121 +106,116 @@ class Porcupine {
...
     }
 
-    if (sensitivities == null || sensitivities.length == 0) {
+    if (sensitivities == null || sensitivities.isEmpty) {
       sensitivities = List.filled(keywordPaths.length, 0.5);
     } else {
       for (double sensitivity in sensitivities) {
         if (sensitivity < 0 || sensitivity > 1 || sensitivity == double.nan) {
-          throw new PvArgumentError(
+          throw PorcupineInvalidArgumentException(
```
laves, line 163: I actually wonder if we should move all validation to native and just have the Flutter plugin be a super thin wrapper...

binding/flutter/lib/porcupine_error.dart
```diff
@@ -9,49 +9,84 @@
 // specific language governing permissions and limitations under the License.
 //
 
-class PvAudioException implements Exception {
+// ignore_for_file: unnecessary_brace_in_string_interps
```
laves, line 12: I think if you just use `$code` instead of `${code}` you don't have to ignore it

binding/flutter/android/src/main/java/ai/picovoice/flutter/porcupine/PorcupinePlugin.java
```diff
@@ -11,23 +11,185 @@
...
+
+        if (!porcupinePool.containsKey(handle)) {
+          result.error(
+                  PorcupineInvalidArgumentException.class.getSimpleName(),
+                  "Invalid Porcupine handle provided to native module.",
+                  null);
+          return;
+        }
+
+        Porcupine porcupine = porcupinePool.get(handle);
```
laves, line 179: You should remove from pool after deleting it

binding/flutter/ios/Classes/SwiftPorcupinePlugin.swift
```diff
@@ -11,9 +11,150 @@
...
+    
+    public static func register(with registrar: FlutterPluginRegistrar) {
+        let instance = SwiftPorcupinePlugin()
+
+        let methodChannel = FlutterMethodChannel(name: "porcupine", binaryMessenger: registrar.messenger())
+        registrar.addMethodCallDelegate(instance, channel: methodChannel)
+    }
+    
+    public func handle(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
+        let method = Method(rawValue: call.method.uppercased()) ?? Method.DEFAULT
```
laves, line 35: again, maybe no default is necessary. Just return an error it you get a weird method call

binding/flutter/ios/Classes/SwiftPorcupinePlugin.swift
```diff
@@ -11,9 +11,150 @@
...
+                    result(errorToFlutterError(PorcupineError.PorcupineInvalidArgumentError("Invalid argument provided to Porcupine 'process'")))
+                }
+            } catch let error as PorcupineError {
+                result(errorToFlutterError(error))
+            } catch {
+                result(errorToFlutterError(PorcupineError.PorcupineInternalError(error.localizedDescription)))
+            }
+            break
+        case .DELETE:
+            let args = call.arguments as! [String: Any]
```
laves, line 109: can you just do this at the top instead of repeating it?

binding/flutter/ios/Classes/SwiftPorcupinePlugin.swift
```diff
@@ -11,9 +11,150 @@
...
+        case .PorcupineActivationRefusedError:
+            return FlutterError(code: "PorcupineActivationRefusedException", message: extractMessage("\(error)"), details: nil)
+        case .PorcupineInternalError:
+            return FlutterError(code: "PorcupineException", message: extractMessage("\(error)"), details: nil)
+        }
+    }
     
-  }
+    private func extractMessage(_ errorMessage: String) -> String {
+        let parts = errorMessage.components(separatedBy: "\"")
```
laves, line 152: Is there not a way to just get the message from the error type? This string parsing seems a bit unnecessary.

binding/flutter/lib/porcupine.dart
```diff
@@ -71,15 +73,27 @@ class Porcupine {
   /// Thows a `PvError` if not initialized correctly
```
laves, line 73: PorcupineException now right?

binding/flutter/lib/porcupine_error.dart
```diff
@@ -9,49 +9,84 @@
...
+
+class PorcupineException implements Exception {
```
laves, line 14: Can you check "Error" vs "Exception" in Flutter? I can't remember, but they were opinionated on which word you use when.

demo/flutter/lib/main.dart
```diff
@@ -53,18 +62,24 @@ class _MyAppState extends State<MyApp> with WidgetsBindingObserver {
 
   @override
   void didChangeAppLifecycleState(AppLifecycleState state) async {
-    if (state == AppLifecycleState.inactive) {
+    if (state == AppLifecycleState.paused) {
```
laves, line 65: I seem to remember there was a specific reason I chose inactive over paused. Make sure you test that putting it in the background of both platforms doesn't mess anything up

demo/flutter/pubspec.yaml
```diff
@@ -11,11 +11,15 @@ dependencies:
...
   
 dev_dependencies:
   flutter_test:
     sdk: flutter
+  lints: ^1.0.1
 
 flutter:
   uses-material-design: true
+  assets:
+    - assets/
```
laves, line 25: what's this for?

laves (changes requested):

binding/flutter/lib/porcupine_error.dart
ksyeo1010: Exception is the normal exceptions we know, i.e. throw and catch. Errors are not supposed to be catched and probably done by programmer's mistake.

demo/flutter/lib/main.dart
ksyeo1010: I remember this messes up the program when mic permission pop up. Right now seems ok going to background and returning back.

binding/flutter/android/src/main/java/ai/picovoice/flutter/porcupine/PorcupinePlugin.java
```diff
@@ -55,55 +53,80 @@ public void onMethodCall(@NonNull MethodCall call, @NonNull Result result) {
...
+          String accessKey = call.argument("accessKey");
+          String modelPath = call.argument("modelPath");
+          ArrayList<String> keywordsList = call.argument("keywords");
+          ArrayList<Double> sensitivitiesList = call.argument("sensitivities");
+
+          Porcupine.BuiltInKeyword[] keywords = null;
+          if (keywordsList != null) {
+            keywords = new Porcupine.BuiltInKeyword[keywordsList.size()];
+            for (int i = 0; i < keywordsList.size(); i++) {
+              keywords[i] = Porcupine.BuiltInKeyword.valueOf(keywordsList.get(i));
```
laves, line 75: we should probably handle the case where we can't translate given keyword to built-in

binding/flutter/lib/porcupine.dart
```diff
@@ -254,27 +176,28 @@ class Porcupine {
...
-    File outputFile = File(outputPath);
-
-    ByteData data =
-        await rootBundle.load(filePath);
-    final buffer = data.buffer;
+    ByteData data;
+    try {
+      data = await rootBundle.load(filePath);
+    } catch (_) {
+      return filePath;
```
laves, line 184: can you just add a comment here about why we do this

binding/flutter/lib/porcupine.dart
```diff
@@ -254,27 +176,28 @@ class Porcupine {
...
-  static String _builtInToKeyword(BuiltInKeyword builtIn) {
-    return builtIn
-        .toString()
-        .split('.')
-        .last
-        .replaceAll('_', ' ')
-        .toLowerCase();
-  }
-
   static Future<String> _extractResource(String filePath) async {
```
laves, line 179: let's change the name of this function to more accurately describe what it's doing. I'm thinking something like "tryExtractFlutterAsset"

demo/flutter/lib/main.dart
```diff
@@ -98,16 +98,15 @@ class _MyAppState extends State<MyApp> with WidgetsBindingObserver {
       });
     } on PorcupineInvalidArgumentException catch (ex) {
       errorCallback(PorcupineInvalidArgumentException(
-          "${ex.message}. Please make sure your accessKey '$accessKey' is a valid access key."));
+          "${ex.message}\nPlease make sure your accessKey '$accessKey' is a valid access key."));
```
laves, line 101: make sure -> ensure  (just slightly more professional sounding)

binding/flutter/ios/Classes/SwiftPorcupinePlugin.swift
```diff
@@ -32,32 +31,61 @@ public class SwiftPorcupinePlugin: NSObject, FlutterPlugin {
...
+                    
+                    var param: [String: Any] = [:]
+                    param["handle"] = handle
+                    param["frameLength"] = Porcupine.frameLength
+                    param["sampleRate"] = Porcupine.sampleRate
+                    param["version"] = Porcupine.version
+                    
+                    result(param)
+                } else {
+                    result(errorToFlutterError(PorcupineError.PorcupineInvalidArgumentError("Invalid argument provided to Porcupine 'from_built_in_keywords'")))
```
laves, line 75: this gets hit if accessKey or keywords are null right? Maybe just be more specific in the throw here and say something like "missing required arguments accessKey and keywords"

laves (changes requested):
It was a long road to get here I know, but this binding is now super light and slick. This pattern is definitely something we're going to use going forward with cross-platform stuff where possible.

laves (approved):

PR #528, opened 2021-11-11
Fix missing python m1

ErisMik:


binding/python/util.py
```diff
@@ -1,5 +1,5 @@
 #
-# Copyright 2020 Picovoice Inc.
+# Copyright 2021 Picovoice Inc.
```
laves, line 2: 2020-2021

laves (changes requested):
can you also add arm64 to the python demo readme?

laves (approved):

PR #529, opened 2021-11-15
v2.0-unity fix

ksyeo1010:
- fix bug where streaming assets would be appended twice on android

### DIFF
--- a/binding/unity/Assets/Porcupine/Porcupine.cs
+++ b/binding/unity/Assets/Porcupine/Porcupine.cs
@@ -408,18 +408,17 @@ namespace Pv.Unity
         private static Dictionary<BuiltInKeyword, string> GetBuiltInKeywordPaths(string platform)
         {
 #if !UNITY_EDITOR && UNITY_ANDROID
-            string keywordFilesDir = Path.Combine(Application.persistentDataPath, "keyword_files", platform);
+            string keywordFilesDir = Path.Combine(Path.Combine(Application.persistentDataPath, "keyword_files"), platform);
             if (!Directory.Exists(keywordFilesDir))
             {
                 Directory.CreateDirectory(keywordFilesDir);
             }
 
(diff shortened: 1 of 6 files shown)


binding/unity/Assets/Porcupine/Porcupine.cs
```diff
@@ -408,18 +408,22 @@ private static string GetDefaultModelPath()
         private static Dictionary<BuiltInKeyword, string> GetBuiltInKeywordPaths(string platform)
         {
 #if !UNITY_EDITOR && UNITY_ANDROID
-            string keywordFilesDir = Path.Combine(Path.Combine(Application.persistentDataPath, "keyword_files"), platform);
+            string keywordFilesDir = Path.Combine(Application.persistentDataPath, "keyword_files", platform);
```
laves, line 411: I know those multi-`Path.Combine`s look weird, but they actually make it compatible with Unity 2017.1 because they conform to an earlier version of C#. Let's keep them there.

binding/unity/Assets/Porcupine/Porcupine.cs
```diff
@@ -444,10 +448,17 @@ private static string GetDefaultModelPath()
...
-            string dstPath = Path.Combine(Application.persistentDataPath, filePath);
-            var loadingRequest = UnityWebRequest.Get(srcPath);
+            if (!filePath.StartsWith(Application.streamingAssetsPath))
+            {
+                throw new PorcupineIOException($"File '{filePath}' not found in streaming assets path.");
+            }
+
+            string dstPath = Path.Combine(
+                Application.persistentDataPath,
+                filePath.Remove(0, Application.streamingAssetsPath.Length + 1));
```
laves, line 458: It would probably be more simple to do a replace operation here. Replace streamingAssetsPath -> persistentDataPath

laves (changes requested):

laves (approved):

PR #530, opened 2021-11-15
ppn-mcu v2.0

mrrostam:
- added keywords for cortexm
- updated static libs
- updated `generate_pv_params_file.py` to generate header files for all languages
- added configurations for de, es, and fr
- updated demos to use new accesskey interface
- updated docs


ErisMik:
For the failing Rust checks in the workflow files replace ` sudo apt install libasound2-dev` with ` sudo apt update && sudo apt install libasound2-dev`. I can do it for you if you'd like.

mrrostam:
> For the failing Rust checks in the workflow files replace ` sudo apt install libasound2-dev` with ` sudo apt update && sudo apt install libasound2-dev`. I can do it for you if you'd like.

I think you fixed this before, right? so I just rebased to get new actions

PR #534, opened 2021-11-18
Web binding update - v2.0

mrrostam:


binding/web/template/src/utils.ts
```diff
@@ -124,7 +124,7 @@ export function getRuntimeEnvironment(): string {
 
  export function isAccessKeyValid(accessKey: string): boolean {
   const accessKeyCleaned = accessKey.trim();
-  if (accessKeyCleaned.length !== 56) { return false; }
+  if (accessKeyCleaned === '') { return false; }
```
ErisMik, line 127: I think in this case we do actually want to check for `length !== 56`. A valid accessKey is of length 56 after all, and this function is not used to fail silently, but throw a loud error which is the original concern.

ErisMik (approved):
Optional comment, otherwise LGTM

binding/web/template/src/utils.ts
mrrostam: Correct! One reason is we do not have this logic in any other binding so it makes the web binding one of a kind which might cause some problems; the other reason is in future we may update the AccessKey length and even have AccessKeys with different sizes. So I think this validation is too limited and fragile to be in a binding. It'd rather to send it to the engine and get some errors back.

PR #537, opened 2021-11-22
V2.0 ios fix

ksyeo1010:
- updated ios bindings, flutter and reactnative to use better error handling method to match android case

### DIFF
--- a/binding/ios/Porcupine.swift
+++ b/binding/ios/Porcupine.swift
@@ -9,49 +9,6 @@
 
 import PvPorcupine
 
-public enum PorcupineError: Error {
-    case PorcupineMemoryError(_ message:String)
-    case PorcupineIOError(_ message:String)
-    case PorcupineInvalidArgumentError(_ message:String)
-    case PorcupineStopIterationError(_ message:String)
-    case PorcupineKeyError(_ message:String)
-    case PorcupineInvalidStateError(_ message:String)
-    case PorcupineRuntimeError(_ message:String)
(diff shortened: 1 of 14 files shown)


binding/ios/Porcupine.swift
```diff
@@ -24,6 +24,34 @@ public enum PorcupineError: Error {
...
+                .PorcupineActivationError(let message),
+                .PorcupineActivationLimitError(let message),
+                .PorcupineActivationThrottledError(let message),
+                .PorcupineActivationRefusedError(let message),
+                .PorcupineError(let message):
+            return NSLocalizedString(message, comment: "")
+        }
+    }
+    
+    public var code : String {
```
laves, line 46: let's make what we're doing here a bit clearer from an API point of view. Since this is only a get property and it's only for us, we should probably just have a utility function called `getPvException` or something that returns the Exception equivalent

binding/ios/Porcupine.swift
```diff
@@ -24,6 +24,34 @@ public enum PorcupineError: Error {
...
+        get {
+            let error = String(describing: self)
+            return error.replacingOccurrences(of: "Error", with: "Exception").components(separatedBy: "(\"")[0]
```
laves, line 49: Is there a way to do this without the `components(separatedBy` chunk? Ideally we'd just get the enum name and replace the last occurrence of the word "Error"

laves (changes requested):

laves (approved):

PR #584, opened 2021-12-07
self-hosted runners for python actions

iliadrm:
added the following self-hosted runners: rpi2, rpi3, rpi4, beaglebone, jetson nano

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -44,7 +44,7 @@ defaults:
     working-directory: binding/python
 
 jobs:
-  build-git-hosted:
+  build:
     runs-on: ${{ matrix.os }}
 
     strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/python.yml
```diff
@@ -59,13 +60,35 @@ jobs:
...
+    - uses: actions/checkout@v2
+
+    - name: Pre-build dependencies
+      run: python3 -m pip install --upgrade pip
+
+    - name: Install dependencies
+      run: pip3 install -r requirements.txt
+
+    - name: Test
+      run: python3 test_porcupine.py ${{secrets.PV_VALID_ACCESS_KEY}}
```
ErisMik, line 94: Is there any reason we can't combine the two jobs seeing as how the steps are nearly incidental? I think it'll be easier to maintain.

iliadrm: I reverted the extra commits.

.github/workflows/python.yml
```diff
@@ -44,11 +44,12 @@ defaults:
     working-directory: binding/python
 
 jobs:
-  build:
-    runs-on: ubuntu-latest
+  build-git-hosted:
```
ErisMik, line 47: Small nitpick, can we call this `build-github-hosted`?

ErisMik (approved):
LGTM after small nitpick

PR #585, opened 2021-12-07
update vue binding to use typescript and added vue types

ksyeo1010:
- vue js to ts
- added types to vue
- added support for vue 2 and vue 3
- changed from renderless component to mixin


demo/vue/src/components/VoiceWidget.vue
```diff
@@ -69,36 +62,44 @@ export default {
...
+    initEngine: function (event: any) {
       this.factoryArgs.accessKey = event.target.value;
       this.isError = false;
       this.isLoaded = false;
       this.isListening = false;
-      this.$refs.porcupine.initEngine();
-    },
+      this.$porcupine.init(
+        this.factoryArgs,
+        PorcupineWorkerFactoryEn,
```
mrrostam, line 81: why not

```suggestion
        this.factory,
```

binding/vue/src/porcupine_types.ts
```diff
@@ -0,0 +1,64 @@
...
+ 
+/**
+* The language-specific worker factory, imported as { PorcupineWorkerFactory } from the 
+* @picovoice/porcupine-web-xx-worker series of packages, where xx is the two-letter language code.
+*/
+export interface PorcupineWorkerFactory extends Object {
+  create: (
+    accessKey: string, 
+    keywords: Array<PorcupineKeyword | string> | PorcupineKeyword | string,
+    keywordDetectionCallback?: CallableFunction,
```
mrrostam, line 42: it has nothing to do with this PR, but i noticed we have this `CallableFunction`s everywhere and I think it's better to make them more specific at some point, something like `(keywordLabel: string) => void`

ksyeo1010: Sure we should do this later on.

binding/vue/README.md
```diff
@@ -12,7 +12,9 @@ applications.
```
mrrostam, line 12: let's update the main readme

binding/vue/README.md
```diff
@@ -12,7 +12,9 @@ applications.
 
 ## Compatibility
```
mrrostam, line 13: # porcupine-web-vue

Renderless Vue component for Porcupine for Web. -> is this still a correct description for this binding?

binding/vue/README.md
```diff
@@ -12,7 +12,9 @@ applications.
...
 
-This library is compatible with Vue 3.
+This library is compatible with Vue:
+- Vue.js 2.6.11+.
+- Vue.js 3.0.0+.
```
laves, line 17:
```suggestion
- Vue.js 2.6.11+
- Vue.js 3.0.0+
```

PR #586, opened 2021-12-08
Added getters/comments to nodejs binding

mrrostam:


binding/nodejs/index.js
```diff
@@ -116,22 +133,56 @@ class Porcupine {
...
+   * @returns the version of the Porcupine engine
+   */
+  get version() {
+    return this._version;
   }
 
+  /**
+   * Process a frame of pcm audio.
+   *
+   * @param {Array} frame 16-bit integers of 16kHz linear PCM mono audio.
```
laves, line 178: We shouldn't mention sample rate since that's a variable.
```suggestion
   * @param {Array} frame of mono, 16-bit, linear-encoded PCM audio.
```

laves (changes requested):

PR #595, opened 2021-12-20
Add support for Java 17 + gradle 7.3

ErisMik:


binding/java/src/ai/picovoice/porcupine/Utils.java
```diff
@@ -160,6 +160,10 @@ private static String getArchitecture() throws RuntimeException {
                 return "x86_64";
             }
         } else if (arch.equals("arm") || arch.equals("aarch64")) {
+            if (ENVIRONMENT_NAME.equals("mac")) {
```
laves, line 163: Can we actually switch this logic around? I think it makes more sense to go:
```java
if (ENVIRONMENT_NAME.equals("mac") {
  // detect and return arch string
} else if ENVIRONMENT_NAME.equals("windows") {
  // detect and return arch string
} else {
  // linux-based environment, detect and return arch string
}
```

laves (changes requested):
Are we now building for Java 17+ only? Ideally, we say Java 17+ for M1, Java 11+ for every other platform. Are we dealing with this split?

ErisMik:
> Are we now building for Java 17+ only? Ideally, we say Java 17+ for M1, Java 11+ for every other platform. Are we dealing with this split?

It' still okay to say Java 11+ as the only requirement. Before it was technically Java 11 -> Java 16 because of Gradle, and now with the Gradle version bump it's Java 11 -> Java 17. The full range of Java versions from 11 -> 17 are also still valid for Mac M1, it's just for versions less than 17 they won't be using native libs, but then again they wouldn't be using a native version of Java either so it's totally okay.

laves (approved):

PR #596, opened 2021-12-20
Add specific error classes for Python

ErisMik:


laves (changes requested):
Can you also add some case-specific error catching to the Python demos like we have with the other SDKs? Mostly just want to show the different activation errors they might encounter on init

ErisMik:
> Can you also add some case-specific error catching to the Python demos like we have with the other SDKs? Mostly just want to show the different activation errors they might encounter on init

Sure. I was planning to have a separate PR for the package release + demo version update for these changes, and I can do that there. Also happy to do it in this PR if you feel that's better.

laves:
Cool, that's fine.

laves (approved):

PR #597, opened 2021-12-21
Release Python package and update demo error logic

ErisMik:


binding/python/__init__.py
```diff
@@ -10,7 +10,18 @@
...
+from .porcupine import PorcupineIOError
+from .porcupine import PorcupineInvalidArgumentError
+from .porcupine import PorcupineStopIterationError
+from .porcupine import PorcupineKeyError
+from .porcupine import PorcupineInvalidStateError
+from .porcupine import PorcupineRuntimeError
+from .porcupine import PorcupineActivationError
+from .porcupine import PorcupineActivationLimitError
+from .porcupine import PorcupineActivationThrottledError
+from .porcupine import PorcupineActivationRefusedError
```
ErisMik, line 24: I guess there was a reason we did the error classes like this in the other packages: Because of how we package the bindings we can't have the unit tests and the package work at the same time. This can be resolved in the future if we want to split the files up like I tried here but the blast radius for that is a bit larger than the scope of this task.

demo/python/porcupine_demo_file.py
```diff
@@ -89,12 +89,32 @@ def main():
...
+    try:
+        porcupine = pvporcupine.create(
+            access_key=args.access_key,
+            library_path=args.library_path,
+            model_path=args.model_path,
+            keyword_paths=keyword_paths,
+            sensitivities=args.sensitivities)
+    except pvporcupine.PorcupineInvalidArgumentError as e:
+        print(f"One or more arguments provided to Porcupine is invalid: {args}")
+        print("If all other arguments seem valid, ensure that 'access_key' is a valid AccessKey")
```
laves, line 101:
```suggestion
        print(f"If all other arguments seem valid, ensure that '{args.access_key}' is a valid AccessKey")
```

laves (changes requested):
Looks good. Seems fine to me that the exceptions live in the main file of the binding. Not much to be gained from splitting it into its own file it seems.

demo/python/porcupine_demo_mic.py
```diff
@@ -110,7 +110,25 @@ def run(self):
                 result = porcupine.process(pcm)
                 if result >= 0:
                     print('[%s] Detected %s' % (str(datetime.now()), keywords[result]))
-
+        except pvporcupine.PorcupineInvalidArgumentError as e:
+            print("One or more arguments provided to Porcupine is invalid")
```
laves, line 114:
```suggestion
            print(f"One or more arguments provided to Porcupine is invalid: {args}")
```

laves (changes requested):

demo/python/porcupine_demo_mic.py
ErisMik: No `args` object in the mic demo unfortunately, we use a class based approach

laves: ah, I see. In that case, ignore this suggestion, but I see that `args.access_key` is used in the following cases, so you should replace those with `'{self._access_key}'`

laves (changes requested):

laves (approved):

PR #605, opened 2021-12-28
Github actions go add platforms and non-English tests

iliadrm:


binding/go/porcupine_test.go
```diff
@@ -13,28 +13,61 @@ package porcupine
...
 	"math"
 	"os"
 	"path/filepath"
 	"testing"
+	"reflect"
 )
 
-var pvTestAccessKey string
+var (
+	pvTestAccessKey string
```
kenarsa, line 26: I know this is not part of this code change but I rather drop the `pv` here. You can use something like `testAccessKey` or even `accessKey`.

binding/go/porcupine_test.go
```diff
@@ -13,28 +13,61 @@ package porcupine
...
+	p               Porcupine
```
kenarsa, line 27: in the previous version, `p` was local in a function and it made sense to have a short name for it. Now that it has a larger scope we prob should give it a better name. `porcupine` ...

binding/go/porcupine_test.go
```diff
@@ -13,28 +13,61 @@ package porcupine
...
-func TestProcess(t *testing.T) {
+func appendLanguage(s string, language string) string {
+	if language == "en" {
+		return s;
+	}
+	return s + "_" + language
+}
+
+func getTestModelPath(language string) string {
+	modelPath, _ := filepath.Abs(fmt.Sprintf(
```
kenarsa, line 46: with nested calls that don't fit into a line please prefer to break them like below:

```go
relPath = fmt.Sprintf(
		"../../lib/common/%s.pv",
		appendLanguage("porcupine_params", language))
filepath.Abs(relPath)
```

binding/go/porcupine_test.go
```diff
@@ -44,16 +77,17 @@ func TestProcess(t *testing.T) {
 	t.Logf("Frame Length: %d", FrameLength)
 	t.Logf("Sample Rate: %d", SampleRate)
 
-	data, err := ioutil.ReadFile(test_file)
+	testFile, _ := filepath.Abs(filepath.Join("../../resources/audio_samples", audioFileName))
```
kenarsa, line 80: `testAudioPath` ?

kenarsa (approved):

binding/go/porcupine_test.go
```diff
@@ -44,16 +79,17 @@ func TestProcess(t *testing.T) {
 	t.Logf("Frame Length: %d", FrameLength)
 	t.Logf("Sample Rate: %d", SampleRate)
 
-	data, err := ioutil.ReadFile(test_file)
+	testAduioPath, _ := filepath.Abs(filepath.Join("../../resources/audio_samples", audioFileName))
```
laves, line 82:
```suggestion
	testAudioPath, _ := filepath.Abs(filepath.Join("../../resources/audio_samples", audioFileName))
```

laves (changes requested):
I'm liking this pattern we've settled on for the new tests!

laves:
Can you look into why Beaglebone failed?

laves (approved):

PR #607, opened 2021-12-29
Github actions dotnet add platforms and non-English tests

iliadrm:


binding/dotnet/PorcupineTest/MainTest.cs
```diff
@@ -33,28 +64,48 @@ public static void ClassInitialize(TestContext testContext)
             }
         }
 
-        [TestMethod]
-        public void TestFrameLength()
+        private static string appendLanguage(string s, string language)
```
laves, line 67: I know this is annoying, but in C# all function names start with an uppercase, regardless of public/private status
```suggestion
        private static string AppendLanguage(string s, string language)
```

binding/dotnet/PorcupineTest/MainTest.cs
```diff
@@ -63,20 +114,50 @@ public void TestProcess()
...
                     results.Add(result);
                 }
             }
 
-            Assert.IsTrue(results.Count == 1 && results[0] == 0, "Expected to find keyword 'porcupine', but no keyword was detected.");
-            p.Dispose();
+            Assert.AreEqual(expectedResults.Count, results.Count, $"Should have found {expectedResults.Count} keywords, but {results.Count} were found.");
+            for (int i = 0; i < results.Count; i++)
+            {
+                Assert.AreEqual(expectedResults[i], results[i], $"Porcupine did not detect keyword #{i+1} correctly.");
```
laves, line 128: Always best to print any relevant info to help with debugging:
```suggestion
                Assert.AreEqual(expectedResults[i], results[i], $"Expected keyword {expectedResults[i]}, but Porcupine detected keyword {results[i]}.");
```

binding/dotnet/PorcupineTest/MainTest.cs
```diff
@@ -90,54 +171,137 @@ public void TestProcessMultiple()
...
+        public void TestSingleKeywordDe()
+        {
+            string language = "de";
+            List<string> keywords = new List<string>() {"heuschrecke"};
+
+            porcupine = Porcupine.FromKeywordPaths(
+                ACCESS_KEY,
+                getKeywordPaths(language, keywords),
+                getModelPath(language),
+                Enumerable.Repeat(0.5f, keywords.Count())
```
laves, line 190: the binding does this by default, so the only reason to pass in values is if we want to use custom (i.e. not 0.5) values

laves (changes requested):

laves (approved):

PR #608, opened 2021-12-29
Github actions C demo add platforms and test file

iliadrm:


demo/c/test_demo.py
```diff
@@ -0,0 +1,225 @@
+#
+# Copyright 2018-2021 Picovoice Inc.
```
laves, line 2: The way we usually do the copyright dates is:
`{year created} - {year last modified}`

In the case of `year created == year last modified` it's just `year created`.
```suggestion
# Copyright 2021 Picovoice Inc.
```

laves (changes requested):
There's quite a bit of repeated code here from the Porcupine Python binding's `utils.py`. Why not just add a `pip install pvporcupine` to the workflow and `from pvporcupine import utils` to this file? Then you can pretty much access anything you need to from the actual test cases.

demo/c/test_demo.py
```diff
@@ -0,0 +1,143 @@
...
+        self.command = ["./build/porcupine_demo_file",
+                    "-l", util.pv_library_path('.'),
+                    "-m", util.pv_model_path('.'),
+                    "-k", "",
+                    "-t", "0.5",
+                    "-w", "",
+                    "-a", sys.argv[1]]
+
+    def test_americano(self):
+        self.command[6] = self.__pv_keyword_path("en", "americano")
```
laves, line 46: instead of using indices (which would be a pain to keep track of over time), you should just use a utility function that takes in named arguments, inserts defaults where appropriate, and returns the command list.

demo/c/test_demo.py
```diff
@@ -0,0 +1,143 @@
...
+        self.command[6] = self.__pv_keyword_path(language, "mon chouchou")
+        self.command[10] = self.__pv_audio_path("mon_chouchou.wav")
+
+        run_demo = subprocess.run(self.command, capture_output=True, text=True)
+
+        self.assertEqual(run_demo.returncode, 0)
+        self.assertTrue(self.__has_detected(run_demo.stdout))
+
+    @staticmethod
+    def __get_environemnt():
```
laves, line 88: I wish there was a way to use only functions from `utils` for stuff like this, but I see now that there isn't any public function available to handle this particular need. We can probably figure out a way to do this in the future updates of the binding so that we don't need this function at all.

demo/c/test_demo.py
```diff
@@ -0,0 +1,143 @@
...
+import subprocess
+import sys
+import unittest
+
+from pvporcupine import util
+
+class PorcupineDemoTestCase(unittest.TestCase):
+    @staticmethod
+    def __has_detected(s):
+        return "detected" in s
```
laves, line 24: we should probably check if it was detected exactly `n` times (`n` being an input argument). If the engine goes off 3 times on a file with only 1 instance of the keyword, we have a problem.

demo/c/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.4)
+cmake_minimum_required(VERSION 3.13)
```
laves, line 1: why was this version increase required?

laves (changes requested):

PR #612, opened 2022-01-05
add indexedDB to web binding

ksyeo1010:
- remove worker messages for file operations
- add indexedDB implementation

Note:
Did not separate logic because it needs a bigger refactoring than I originally expected.
When initializing WASM, it grabs the exported functions and also imports the js functions needed in wasm at the same time. The imported function use the `aligned_alloc` function which is exported having a cyclic reference. This works when everything is in the same function but when I tried to refactor,  the `aligned_alloc` function is undefined. PM me for more dails if needed. If anyone has a solution, I'll be happy to add it here.

### DIFF
--- a/binding/web/template/src/porcupine.ts
+++ b/binding/web/template/src/porcupine.ts
@@ -39,8 +39,6 @@ import {
 const DEFAULT_SENSITIVITY = 0.5;
 const PV_STATUS_SUCCESS = 10000;
 
-type EmptyPromise = (value: void) => void;
-
 /**
  * JavaScript/WebAssembly Binding for the Picovoice Porcupine wake word engine.
  *
--- a/binding/web/template/src/porcupine.ts
+++ b/binding/web/template/src/porcupine.ts
@@ -85,8 +85,6 @@ export class Porcupine implements PorcupineEngine {
 
   private static _porcupineMutex = new Mutex;
 
-  private static _storage = getPvStorage();
-
   private constructor(handleWasm: PorcupineWasmOutput, keywordLabels: ArrayLike<string>) {
     Porcupine._frameLength = handleWasm.frameLength;
     Porcupine._sampleRate = handleWasm.sampleRate;
@@ -323,6 +321,8 @@ export class Porcupine implements PorcupineEngine {
     const memoryBufferUint8 = new Uint8Array(memory.buffer);
     const memoryBufferInt32 = new Int32Array(memory.buffer);
     const memoryBufferFloat32 = new Float32Array(memory.buffer);
+
+    const storage = getPvStorage();
     
     const pvConsoleLogWasm = function (index: number): void {
       // eslint-disable-next-line no-console
@@ -439,7 +439,7 @@ export class Porcupine implements PorcupineEngine {
     ): Promise<void> {
       const path = arrayBufferToStringAtIndex(memoryBufferUint8, pathAddress);
       try {
-        const contentBase64 = await Porcupine._storage.getItem(path);
+        const contentBase64 = await storage.getItem(path);
         const contentBuffer = base64ToUint8Array(contentBase64);
         // eslint-disable-next-line
         const contentAddress = await aligned_alloc(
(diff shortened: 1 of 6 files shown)


binding/web/template/src/porcupine.ts
```diff
@@ -423,11 +342,11 @@ export class Porcupine implements PorcupineEngine {
         throw new Error(`assertion failed at line ${line} in "${fileName}"`);
       }
     };
-
+  
     const pvTimeWasm = function (): number {
       return Date.now() / 1000;
     };
-
+
```
ErisMik, line 349: What are these whitespace changes about?

binding/web/template/src/porcupine.ts
```diff
@@ -609,10 +516,7 @@ export class Porcupine implements PorcupineEngine {
     ): Promise<void> {
       const path = arrayBufferToStringAtIndex(memoryBufferUint8, pathAddress);
       try {
-        await pvFileOperationHelper({
-          command: 'file-delete',
-          path: path,
-        });
+        await storage.removeItem(path);
```
ErisMik, line 519: So according to the types, `removeItem()` could be `async` (indexedDB) or not (localstorage)? I'm not too familiar with Javascript but how will that work out here with the await when it is not async?

ksyeo1010: It basically wraps it into a promise which in this scenario doesn't really affect the function.

ErisMik: Sounds good :+1:

ErisMik (approved):

binding/web/template/src/porcupine.ts
```diff
@@ -565,11 +479,7 @@ export class Porcupine implements PorcupineEngine {
         contentAddress
       );
       try {
-        await pvFileOperationHelper({
-          command: 'file-save',
-          path: path,
-          content: content,
-        });
+        await storage.setItem(path, content);
```
laves, line 482: pedantic, but I think `setItem` should be `putItem` since we're talking about storage and that's what the db transaction is called

binding/web/template/src/porcupine_worker.ts
```diff
@@ -78,55 +78,23 @@ onmessage = function (
   event: MessageEvent<PorcupineWorkerRequest>
 ): void {
   switch (event.data.command) {
-    case 'file-save-succeeded':
-      Porcupine.resolveFilePromise(event.data.message);
-      Porcupine.clearFilePromises();
+    case 'init':
+      init(event.data.accessKey, event.data.keywords,event.data.start);
```
laves, line 82:
```suggestion
      init(event.data.accessKey, event.data.keywords, event.data.start);
```

binding/web/template/src/porcupine_types.ts
```diff
@@ -82,18 +67,11 @@ export type PorcupineWorkerResponseError = {
...
-  content?: string;
-};
-
 export type PorcupineWorkerResponse =
   | PorcupineWorkerResponseReady
   | PorcupineWorkerResponseFailed
   | PorcupineWorkerResponseKeyword
-  | PorcupineWorkerResponseError
-  | PorcupineWorkerResponseFileOperation;
+  | PorcupineWorkerResponseError;
```
laves, line 74: why do we have a ResponseError and a ResponseFailed? Is there any way we can just have one type of error response?

laves (changes requested):

binding/web/template/src/porcupine.ts
ErisMik: As I understood it was done this way to match the `localStorage` API so it could be a drop-in fallback replacement

laves: ah okay, that makes sense

binding/web/template/src/porcupine_types.ts
ksyeo1010: I think it was initially here to create a distinction for errors when you create factory and process audio frames. So the porcupine worker factory is a async call that returns a promise, which is resolved if worker `init` succeeded, rejected if `failed` is the response message. Lines 40-44 in porcupine_worker_factory.ts basically explain this behaviour:
```javascript
// n.b. The *worker* creation is itself synchronous. But, inside the worker is an async
// method of PorcupineFactory which is initializing. This means the worker is not actually ready
// for voice processing immediately after instantiation. When its initialization completes,
// we receive a special PorcupineWorkerMessageOut message and resolve the worker promise.
```

laves: okay, that's a bit weird, but I guess it's fine

laves (approved):

PR #616, opened 2022-01-10
C actions

ksyeo1010:
- added c demo actions
- c demo micbuild and filebuild
- run file build with multiple languages
- change cmake min version to 3.13 as per https://cmake.org/cmake/help/latest/release/3.13.html


laves (changes requested):
input args to c-filedemo.yml should just be the input args to the filedemo. That way we can run any combinations of files through it.

laves (approved):

PR #622, opened 2022-01-12
web core

ksyeo1010:
- create separate package to hold porcupine types


ErisMik:
Is this just a copy of the types from the standard binding or should the standard binding have this as a dependency?

ksyeo1010:
> Is this just a copy of the types from the standard binding or should the standard binding have this as a dependency?

The standard binding should have this as a dependency.

ErisMik:
> > Is this just a copy of the types from the standard binding or should the standard binding have this as a dependency?
> 
> The standard binding should have this as a dependency.

Okay, I don't see any of that here, are we waiting to push the package out first?

ErisMik (approved):

ksyeo1010:
> > > Is this just a copy of the types from the standard binding or should the standard binding have this as a dependency?
> > 
> > 
> > The standard binding should have this as a dependency.
> 
> Okay, I don't see any of that here, are we waiting to push the package out first?

yes, I am gonna make a separate one for it. going to do type packages first

laves (approved):

binding/web/core/README.md
```diff
@@ -0,0 +1,16 @@
+# porcupine-web-core
+
+Package containing types for Porcupine web bindings. 
+
+For more details, check [binding/web](binding/web).
```
kenarsa, line 5: I think the link won't work. can you give it a try?

ksyeo1010: I removed to link since it was in a different directory at start. Now that is is a subdir from the web binding I don't think we need it.

PR #644, opened 2022-01-20
V2.1 unity

mrrostam:
- added m1 support
- removed "checkArchitecture"
- released two packages: one with m1 lib and the other without so that people still can open the package using the old versions of editor without extra confusing steps

@laves one question: I checked the previous packages as well and noticed that the demo is included inside the package. is it expected?


laves (changes requested):
@mrrostam can you add some info to the README regarding our compatibility? Basically, we now support all platforms except M1 on Unity 2017.4 using one package and all platforms including M1 on Unity 2021.2 using the other

PR #684, opened 2022-03-11
added non-en tests to rust action

mrrostam:
- refactored the test and added non-en cases
- setup and added self-hosted devices


ErisMik (approved):
Can you do me a favour and run `cargo fmt` in the binding dir to format the code before you merge?

PR #690, opened 2022-03-18
add speed test

ksyeo1010:
- add speed test to c, java, node.js, android and ios bindings


laves (changes requested):
Pedantic, but you call this variable `threshold` everywhere. I think it's a bit too generic. Can you do something like `performance_threshold_sec` or something? That way we know a) what the threshold is for and b) what the unit is.

laves (approved):

PR #694, opened 2022-03-29
Added "start" button to web demos

mrrostam:


ksyeo1010:
all good except for react demo can we make it so 'invalid access key' error doesnt flash in the screen?

Issue #662, opened 2022-02-03
Porcupine Issue: Respeaker-rpi0/porcupine_demo_mic.c demo doesn't work on Raspberry OS Bullseye


hedgemybets:
### Expected behaviour

The instructions in the README are to first install wiringpi:
sudo apt-get install wiringpi.

Expect behavior is for the library to install without errors.

### Actual behaviour

E: Package 'wiringpi' has no installation candidate

In researching this, it appears that if your RPi OS has the gpio command, wiringpi is already installed and you're in luck.
If you have Bullseye, gpio is not installed since wiringpi has been deprecated.

I see two ways to fix this. One is to use another library to communicate via SPI, such as [spidev-lib](https://github.com/milekium/spidev-lib). The demo C program would need to be modified.
The other is to build and install wiringpi from the clone of the git repo of [wiringpi](https://github.com/WiringPi/WiringPi).

kenarsa:
I think it works on Raspbian. If you can reproduce we can take a look. Otherwise, feel free to open issues on those repos if you have not done already.

hedgemybets:
Thanks, it probably does work on Raspbian per my research, however it doesn't work on Raspberry Pi OS (rebranded name of Raspbian) Bullseye, the latest release, since the library required is deprecated and not available. Maybe it's more of a documentation issue to specify the demo doesn't work on Bullseye or provide a workaround.

kenarsa:
@hedgemybets thank you for the information. this is super helpful. We do want to be able to run on the latest official OS anyway. So we need to fix this. PRs are really welcome as I think it would be hard for me to prioritize this near-term. If you find a solution I appreciate a PR :)

hedgemybets:
Thanks, FYI, I did compile the WiringPi library without error on my Pi Zero W, but I seem to have an issue with my GCC toolchain as I get a segmentation fault when running the "gpio -v" command or linking/running the porcupine demo program.

My workaround will be to take the easy way out and downgrade the O/S and try again. As for the bug tag, maybe "enhancement request" is more accurate. I believe there are probably a lot of libraries and dependencies that haven't caught up with the recent release of Bullseye.

PR #697, opened 2022-03-31
Remove WiringPI from respeaker-rpi0 demo

ErisMik:
Fixes https://github.com/Picovoice/porcupine/issues/662 :)


demo/respeaker-rpi0/porcupine_demo_mic.c
```diff
@@ -33,10 +33,56 @@ static const uint8_t YELLOW_RGB[3] = {255, 255, 51};
 
 static volatile bool is_interrupted = false;
 
+static const uint8_t spi_mode = 0;
```
mrrostam, line 36: for future reference, can you add some kind of source for these values?

mrrostam (approved):

PR #727, opened 2022-06-09
.NET 6

ErisMik:


binding/dotnet/Porcupine/Porcupine.csproj
```diff
@@ -72,21 +77,23 @@ Porcupine is:
...
-                build/netcoreapp3.1/lib/beaglebone/libpv_porcupine.so
+                build/netcoreapp3.1/lib/beaglebone/libpv_porcupine.so;
+                build/net6.0/lib/beaglebone/libpv_porcupine.so;
             </PackagePath>
             <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
             <Link>lib\beaglebone\libpv_porcupine.so</Link>
             <Visible>false</Visible>
         </Content>
-		<Content Include="..\..\..\lib\mac\x86_64\libpv_porcupine.dylib">
+		<Content Include="..\..\..\lib\mac\**\*" Exclude="..\..\..\lib\mac\arm11\*">
```
laves, line 87: don't think this arm11 exclude needs to be here

binding/dotnet/Porcupine/Utils.cs
```diff
@@ -59,6 +59,10 @@ public static string PvLibraryPath(string libName)
             {
                 return Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), $"lib/{_env}/x86_64/{libName}.dylib");
             }
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX) && _isArm)
```
laves, line 62: I thought you were also going to check if they were running on dotnet 6.0 when loading M1 lib?

laves (changes requested):

laves:
Just wondering - did you test the package in a .NET 5.0 app?

binding/dotnet/Porcupine/Utils.cs
ErisMik: As far as I can tell it detects the arch of the runtime, not the platform - so we are in the clear :)

ErisMik:
> Just wondering - did you test the package in a .NET 5.0 app?

Just did, it appears to run OK

PR #733, opened 2022-06-28
Python release

mrrostam:
parameterized the test cases, so now it's easier to import the test data from somewhere else.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -62,6 +62,9 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
 
+    - name: Install dependencies
+      run: pip install -r requirements.txt
+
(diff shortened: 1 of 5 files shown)


ErisMik (approved):

laves (commented):
Just a question - why did we use this over the built-in unittest sub test? https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests

mrrostam:
> Just a question - why did we use this over the built-in unittest sub test? https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests

I didn't know `subtest` can do basically the same thing, although i think `parameterized` does it in a cleaner fashion; do you want me to replace it?

PR #738, opened 2022-06-29
Rust additional language support

ErisMik:


binding/rust/tests/porcupine_tests.rs
```diff
@@ -255,6 +255,46 @@ mod tests {
         );
     }
 
+    #[test]
```
mrrostam, line 258: is there a rusty way to parameterized these tests? it helps with reducing the the # of lines we need to copy/paste after adding new languages; also the ultimate goal is to take the test data out of the test files and put it somewhere so that each binding can read it and generate necessary tests.

ErisMik: No built in way but the defacto way is to use macros, I can go ahead and try that

mrrostam (approved):

PR #780, opened 2022-09-02
tk demo update

kenarsa:


demo/tkinter/main.py
```diff
@@ -40,7 +39,8 @@ def run(self):
         recorder = None
```
mrrostam, line 40: what do you thing about adding a label here to give some clue to users? something like

```
tk.Label(window, text='Listening for a wake word:').pack()
```

kenarsa: I rather keep it as is to match the YouTube video. Great idea for the next demo though

PR #781, opened 2022-09-02
Angular update

ksyeo1010:
- update angular version to 13

Will fix demos after approve


mrrostam:
do we want `demo/angular/src/lib/porcupine_params.js` to be inside the demo folder? why not generate it as a setup step?

ksyeo1010:
> do we want `demo/angular/src/lib/porcupine_params.js` to be inside the demo folder? why not generate it as a setup step?

yes, forgot to gitignore

mrrostam (approved):

mrrostam (approved):

PR #786, opened 2022-09-09
Link fixes

kenarsa:


README.md
```diff
@@ -297,7 +297,7 @@ npm install
 npm run start
 ```
 
-Open http://localhost:5000 in your browser to try the demo.
+Open `http://localhost:5000` in your browser to try the demo.
```
ksyeo1010, line 300: should we actually link them [http://localhost:5000](http://localhost:5000) and then add a next line disable?

kenarsa: I think we went with this model in other repos. Let's keep doing this.

ksyeo1010 (approved):

PR #833, opened 2022-12-07
go codestyle

ksyeo1010:


.github/workflows/go-codestyle.yml
```diff
@@ -0,0 +1,63 @@
+name: Go
```
laves, line 1: Go Codestyle

laves (changes requested):

laves:
Also, now that I see the trend - can you break Flutter lint out into its own file? Seems weird that that job is combined with build/test.

ksyeo1010:
> Also, now that I see the trend - can you break Flutter lint out into its own file? Seems weird that that job is combined with build/test.

Some are together some are separate (js and flutter). I can separate them all.

PR #837, opened 2022-12-08
update workflow

ksyeo1010:


.github/workflows/js-codestyle.yml
```diff
@@ -0,0 +1,145 @@
+name: JavaScript Codestyle
```
laves, line 1: The only problem with this monofile approach for JS is that it seems that it's going to run on every single js file in the repo after I change just a single SDK. I wonder if there's a way to filter for each job so that if I, for instance, change a JS file in React SDK, only the React codestyle job runs?

laves (changes requested):

.github/workflows/js-codestyle.yml
ksyeo1010: I will have to do one file per framework then.

laves: Okay, yeah, let's just do a file per then.

laves (approved):

PR #850, opened 2022-12-21
Migrate setup-java action to use Temurin

gdams:
AdoptOpenJDK has moved the the Eclipse Foundation and now distributes binaries under the new name "Eclipse Temurin"


ErisMik:
Thanks for the PR!

PR #860, opened 2023-01-07
Flutter test with test_data file

ErisMik:


demo/flutter/integration_test/app_test.dart
```diff
@@ -0,0 +1,128 @@
...
+import 'package:wav/wav.dart';
+
+import 'package:porcupine_flutter/porcupine.dart';
+import 'package:porcupine_flutter/porcupine_error.dart';
+
+void main() {
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  final String accessKey = "{TESTING_ACCESS_KEY_HERE}";
+  final String platform = Platform.isAndroid ? "android" : Platform.isIOS ? "ios" : "error";
```
laves, line 18: I think we should fail here if not running on android or ios

demo/flutter/integration_test/app_test.dart
```diff
@@ -0,0 +1,128 @@
...
+
+      var audioFileData = await rootBundle.load(audioPath);
+      Wav audioFile = await Wav.read(audioFileData.buffer.asUint8List());
+      List<int> pcm = audioFile.channels[0].map((f) {
+          var i = (f * INT16_MAX).truncate();
+          if (f > INT16_MAX) i = INT16_MAX;
+          if (f < INT16_MIN) i = INT16_MIN;
+          return i;
+        }).toList();
+    return pcm;
```
laves, line 32: why don't we load the wav file as a 16-bit pcm?  The wav dart package claims that it can read 16-bit ints

demo/flutter/integration_test/app_test.dart
```diff
@@ -0,0 +1,128 @@
...
+          porcupine.delete();
+          expect(detections.length, equals(1), reason: "Number of detections for ${language} ${keyword} was incorrect");
+          expect(detections[0], equals(0), reason: "Porcupine returned wrong keyword index for ${language} ${keyword}");
+        }
+      });
+
+    testWidgets('Test multipleKeyword all languages',
+      (tester) async {
+        String testDataJson = await rootBundle.loadString('assets/test_resources/test_data.json');
+        var testData = json.decode(testDataJson);
```
laves, line 83: since this is used across multiple tests, we should load it once at the top/in a setup op. If that's not possible, a utility function will do.

demo/flutter/integration_test/app_test.dart
```diff
@@ -0,0 +1,128 @@
...
+              modelPath: modelPath);
+          } on PorcupineException catch (ex) {
+            expect(ex, equals(null), reason: "Failed to initialize Porcupine for ${language}: ${ex}");
+            return;
+          }
+
+          String audioPath = "assets/test_resources/audio_samples/multiple_keywords_$language.wav";
+          if (language == "en") {
+            audioPath = "assets/test_resources/audio_samples/multiple_keywords.wav";
+          }
```
laves, line 110: not that it's a big deal, but you can do this in a one liner:
```dart
String audioPath = "assets/test_resources/audio_samples/multiple_keywords${language != "en" ? language : ""}.wav";
```

laves (changes requested):

demo/flutter/integration_test/app_test.dart
ErisMik: I think that it meant the library can read WAV files that arestored in the Int16 format. The package only presents the one interface to get the PCM and it's in Float32 format.

laves: ah, I see. That's unfortunate.

laves (approved):

PR #861, opened 2023-01-07
add react tests

ksyeo1010:


binding/react/test/use_porcupine.test.ts
```diff
@@ -0,0 +1,106 @@
...
+        BuiltInKeyword.Porcupine,
+        { publicPath: "/test/porcupine_params.pv", forceWrite: true }
+      )
+    ).then(() => {
+      expect(result.current.isLoaded).to.be.false;
+      expect(result.current.error).to.contain("Invalid AccessKey");
+    });
+  });
+
+  it('should be able to process audio', () => {
```
ErisMik, line 77: I guess we did decide that we weren't going to do all the languages via the `test_data.json` file for the UI frameworks right? But it doesn't look all that difficult, so could we add a task to tracking to do that?

ErisMik (approved):

binding/react/test/use_porcupine.test.ts
ksyeo1010: hmm I guess I can do it as part of this PR

binding/web/scripts/copy_ppn.js
```diff
@@ -1,23 +1,6 @@
...
-  "grasshopper_wasm.ppn",
-  "hey google_wasm.ppn",
-  "hey siri_wasm.ppn",
-  "jarvis_wasm.ppn",
-  "ok google_wasm.ppn",
-  "picovoice_wasm.ppn",
-  "porcupine_wasm.ppn",
-  "terminator_wasm.ppn",
-]
-
```
laves, line 20: when I run `copyppn` and `setup-test` it copies all the keywords from the wasm folder. We don't want to release most of them as built-ins and our tests don't really need most of them either. Can you change the script to only copy over this subset of the `_wasm.ppn` files?

laves (changes requested):

laves (changes requested):

binding/web/scripts/setup_test.js
```diff
@@ -51,10 +42,6 @@ try {
     fs.copyFileSync(join(paramsSourceDirectory, file), join(outputDirectory, file));
   });
 
-  fs.readdirSync(paramsSourceDirectory).forEach(file => {
-    fs.copyFileSync(join(paramsSourceDirectory, file), join(outputDirectory, file));
-  });
-
   fs.copyFileSync(testDataSource, join(outputDirectory, 'test_data.json'));
 
   fs.mkdirSync(join(outputDirectory, 'keywords'), { recursive: true });
```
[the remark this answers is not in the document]
ErisMik: I think the idea is to dumb copy all the keyword files, since when additional tests are added to `test_data.json` we don't want to have to go in and update all these copy scripts

ErisMik (approved):

laves (approved):

PR #880, opened 2023-01-18
RN Tests

ErisMik:


.github/workflows/react-native-tests.yml
```diff
@@ -0,0 +1,78 @@
...
+  pull_request:
+    branches: [ master ]
+    paths:
+      - 'binding/react-native/test-app/**'
+      - 'lib/common/**'
+      - 'lib/android/**'
+      - 'lib/ios/**'
+      - 'resources/keyword_files/android/**'
+      - 'resources/keyword_files/ios/**'
+      - '.github/workflows/react-native-tests.yml'
```
laves, line 23: don't we want these to run whenever there's a change to `binding/react-native/**` ?

laves (changes requested):

.github/workflows/react-native-tests.yml
ErisMik: Ah, right!

laves (approved):

PR #890, opened 2023-01-31
C tests from JSON

laves:


.github/workflows/dotnet.yml
```diff
@@ -14,8 +14,14 @@ on:
...
       - 'lib/windows/**'
-      - 'resources/**'
-      - '!resources/lint/**'
+      - 'resources/audio_samples/**'
+      - 'resources/keyword_files/beaglebone/**'
+      - 'resources/keyword_files/jetson/**'
+      - 'resources/keyword_files/linux/**'
+      - 'resources/keyword_files/mac/**'
+      - 'resources/keyword_files/raspberry-pi/**'
+      - 'resources/keyword_files/windows/**'
```
ErisMik, line 23: I don't think this does what we want it to do since it only covers the `en` case right? We should switch to something like this (assuming actions supports the syntax) for all the workflows:
```suggestion
      - 'resources/keyword_files*/beaglebone/**'
      - 'resources/keyword_files*/jetson/**'
      - 'resources/keyword_files*/linux/**'
      - 'resources/keyword_files*/mac/**'
      - 'resources/keyword_files*/raspberry-pi/**'
      - 'resources/keyword_files*/windows/**'
```

laves: good point!

.github/workflows/c-demos.yml
```diff
@@ -36,12 +36,12 @@ on:
       - 'lib/raspberry-pi/**'
       - 'lib/windows/**'
       - 'resources/audio_samples/**'
-      - 'resources/keyword_files/beaglebone/**'
-      - 'resources/keyword_files/jetson/**'
-      - 'resources/keyword_files/linux/**'
-      - 'resources/keyword_files/mac/**'
-      - 'resources/keyword_files/raspberry-pi/**'
-      - 'resources/keyword_files/windows/**'
+      - 'resources/keyword_files*/beaglebone/**'
```
ErisMik, line 39: I think you forgot to add it to the `push` section above

PR #913, opened 2023-03-30
Porcupine - v2.2 - iOS

ErisMik:


binding/ios/Porcupine.swift
```diff
@@ -92,7 +92,7 @@ public class Porcupine {
         }
 
         var keywordPathsArgs = keywordPaths
-        for i in 0..<keywordPathsArgs.count where !FileManager().fileExists(atPath: keywordPathsArgs[i] {
+        for i in 0..<keywordPathsArgs.count where !FileManager().fileExists(atPath: keywordPathsArgs[i]) {
```
mrrostam, line 95: How did it function previously with this error?

mrrostam (approved):

binding/ios/Porcupine.swift
ErisMik: I don't think it did. This broke as part of a codestyle change that went in AFTER the last release and I guess the pipeline never ran.

laves (approved):

PR #914, opened 2023-03-30
Porcupine - v2.2 - android

ErisMik:


.github/workflows/android-appcenter.yml
```diff
@@ -5,13 +5,19 @@ on:
   push:
     branches: [ master ]
     paths:
-      - 'demo/android/Activity/porcupine-activity-demo-app/build.gradle'
       - '.github/workflows/android-appcenter.yml'
+      - 'demo/android/Activity/porcupine-activity-demo-app/**'
+      - '!demo/android/Activity/porcupine-activity-demo-app/README.md'
```
laves, line 10: there's not a README at that location. I think the paths should be `demo/android/Activity/**` and `!demo/android/Activity/README.md`

.github/workflows/android-appcenter.yml
```diff
@@ -32,7 +38,7 @@ jobs:
       run: npm install -g appcenter-cli
 
     - name: set up JDK 11
-      uses: actions/setup-java@v2
+      uses: actions/setup-java@v3
       with:
         java-version: '11'
         distribution: 'adopt'
```
laves, line 44: should be `temurin` right?

binding/android/Porcupine/porcupine/build.gradle
```diff
@@ -34,3 +33,10 @@ if (file("${rootDir}/publish-mavencentral.gradle").exists()) {
 dependencies {
     implementation fileTree(dir: "libs", include: ["*.jar"])
 }
+
+task copyLibs(type: Copy) {
+    from("${rootDir}/../../lib/android")
+    into("${rootDir}/porcupine/src/main/jniLibs")
+}
+
+preBuild.dependsOn copyLibs
```
laves, line 42: I think it's time we move to 100% dependency on the gradle copy for our libs and models. Delete all models, built-in keywords and libs from the project, add `.gitkeeps` to the folders, add them to `.gitignore` and finally add the copies to this gradle script. That way we never need to remember to keep them up to date - they just always will be up to date

demo/android/Activity/copy_test_resources.sh
```diff
@@ -1,34 +1,38 @@
-if [ ! -d "./porcupine-activity-demo-app/src/androidTest/assets/test_resources/audio_samples" ]
+LIB_DIR="../../../lib"
```
laves, line 1: we can also add the contents of this script to the gradle build - though not as important as the binding

.github/workflows/android-perf.yml
laves, line 1: I'm not sure the triggers on this file make sense. If we update `lib/android`, this test will run, but using the android lib from maven - right? If we want to trigger like that, we'll have to use a local build of the android lib. Otherwise we can wait for the `build.gradle` to update, meaning that a new copy of the lib has been staged on Maven.

laves (changes requested):

demo/android/Activity/copy_test_resources.sh
ErisMik: This one is a bit hard to do in gradle because of the per-language dirs and the re-organization on the destination. I think it's possible but it's a fair amount of work/research for release. Okay if I make a ticket to migrate this one later?

ksyeo1010 (approved):

laves (approved):

PR #927, opened 2023-04-06
Porcupine - v2.2 - Java

ErisMik:


.github/workflows/java-perf.yml
```diff
@@ -91,10 +91,10 @@ jobs:
           performance_threshold_sec: 2.0
         - machine: beaglebone
           num_test_iterations: 10
-          performance_threshold_sec: 12.0
+          performance_threshold_sec: 20.0
```
laves, line 94: that's a pretty big change in beaglebone - can we account for that?

ErisMik: Running some tests on Beaglebones now to see if it's the device or something else, just slow because well, Beaglebone. It is odd that it's Porcupine only

ErisMik: @laves We had two Beaglebones for GitHub actions, looks like the first one died and the second has different performance characteristics (swap/sd card/java version). Hard to tell which specific thing since the first one is fubar.

laves (approved):

PR #933, opened 2023-04-11
v2.2 web

ksyeo1010:


binding/angular/projects/porcupine-angular/src/lib/porcupine.service.ts
```diff
@@ -30,7 +30,7 @@ export class PorcupineService implements OnDestroy {
 
   public isLoaded$: Subject<boolean> = new Subject<boolean>();
   public isListening$: Subject<boolean> = new Subject<boolean>();
-  public error$: Subject<string | null> = new Subject<string | null>();
+  public error$: Subject<Error | string | null> = new Subject<Error | string | null>();
```
laves, line 33: Is this ever going to be just a string anymore or are they all Errors now?

ksyeo1010: processcallback returns string

laves: is it a big deal to change that as well?

PR #976, opened 2023-05-16
add multilingual react demos

ksyeo1010:


laves (changes requested):
The code looks good, but we should make the demo the same as the web/mobile demos where you pick a keyword to detect from the list of available keywords rather than init them all at once.

laves (approved):

PR #1017, opened 2023-06-23
demo audit

ksyeo1010:


.github/workflows/android-demos.yml
```diff
@@ -0,0 +1,102 @@
...
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Set up Node.js LTS
+      uses: actions/setup-node@v3
+      with:
+        node-version: lts/*
+
+    - name: Install AppCenter CLI
+      run: npm install -g appcenter-cli
```
laves, line 93: don't need node or app center if we're not uploading the APK

.github/workflows/angular-demos.yml
```diff
@@ -46,3 +50,30 @@ jobs:
...
+  
+  build-stt-demo:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: demo/angular-stt
+
+    strategy:
+      matrix:
+        node-version: [lts/*]
```
laves, line 62: don't need the matrix if you're just doing LTS

laves (changes requested):

laves (approved):

PR #1070, opened 2023-08-08
update unity binding with latest uvp

ksyeo1010:


binding/unity/Assets/Porcupine/PorcupineManager.cs
```diff
@@ -107,56 +105,51 @@ private void OnFrameCaptured(short[] pcm)
...
+            if (VoiceProcessor.Instance.IsRecording)
             {
-                if (_voiceProcessor.IsRecording)
-                {
-                    _voiceProcessor.StopRecording();
-                }
-
-                _voiceProcessor.OnFrameCaptured -= OnFrameCaptured;
-                _voiceProcessor = null;
+                VoiceProcessor.Instance.StopRecording();
```
laves, line 152: the if should be if `numFrameListeners == 0` we call stop. If people are using separate managers, we only want to call stop if there are no other managers actively listening to the stream.

binding/unity/copy.sh
```diff
@@ -27,6 +27,13 @@ cp -rp ../../resources/keyword_files/linux/* ./Assets/StreamingAssets/keyword_fi
 cp -rp ../../resources/keyword_files/mac/* ./Assets/StreamingAssets/keyword_files/mac
 cp -rp ../../resources/keyword_files/windows/* ./Assets/StreamingAssets/keyword_files/windows
 
+echo "Copying Voice Processor files..."
+if [ ! -d "./Assets/Porcupine/VoiceProcessor" ]
+then
+    mkdir -p ./Assets/Porcupine/VoiceProcessor
+fi
+cp -rp VoiceProcessor/Assets/UnityVoiceProcessor/*.cs ./Assets/Porcupine/VoiceProcessor
```
laves, line 35: I can't comment below but two things:
1) delete the 2.2.1 unity package as we're only keeping the latest of each minor version
2) why did we add 2.75 MB to unity packages - did we add that much code? Can you double-check that there isn't test data or code in the packages.

laves (changes requested):

binding/unity/copy.sh
ksyeo1010: https://github.com/Picovoice/porcupine/blob/master/binding/unity/porcupine-2.2.1.unitypackage, package size doesnt seem that different

ksyeo1010: I also checked previous package before exporting the new package and added everything the one before had. the previous package did have test code in it so I didn't really exclude it in the latest one. Do you want to exclude that?

laves: right, 2.75MB is the total size. I do think we should exclude the test code since it doesn't work at all without the test assets. When someone installs it into their project, I highly doubt they're going to import a bunch of our test assets to run the test code.

laves (approved):

PR #1101, opened 2023-10-06
v3.0 unity

ksyeo1010:


binding/unity/Assets/Porcupine/Porcupine.cs
```diff
@@ -368,6 +383,26 @@ public void Dispose()
             Dispose();
         }
 
+        private string[] GetMessageStack()
+        {
+            int messageStackDepth;
+            IntPtr messageStackRef;
+
+            pv_get_error_stack(out messageStackRef, out messageStackDepth);
```
laves, line 391: check the return code

binding/unity/Assets/Porcupine/Porcupine.cs
```diff
@@ -317,34 +330,36 @@ public int Process(short[] pcm)
         /// <returns>.NET exception</returns>
         private static PorcupineException PorcupineStatusToException(
             PorcupineStatus status,
-            string message = "")
+            string message = "",
```
laves, line 333: gotta update the doc comment above

laves (changes requested):

PR #1110, opened 2023-10-13
v3.0 rn

laves:
Because we have to send the errors over the bridge for Flutter and React Native, I opted to just send the full error message rather than trying to serialize/deserialize the stack to send over the bridge.


ksyeo1010 (approved):
I think it is a good time to run yarn upgrade so patch versions get upgraded at least.

PR #1111, opened 2023-10-16
v3.0 web fix

ksyeo1010:


binding/web/src/porcupine.ts
```diff
@@ -621,6 +621,7 @@ export class Porcupine {
       pvPorcupineDelete: pv_porcupine_delete,
       pvPorcupineProcess: pv_porcupine_process,
       pvGetErrorStack: pv_get_error_stack,
+      pvFreeErrorStack: pv_free_error_stack,
```
laves, line 624: how was this passing the tests before? Aren't we calling free when we get the error stack? or is that not something we have a test for?

ksyeo1010: so aside from the init function, process is the only one using it. init doesnt need the function cause it is in the same function

ksyeo1010: process doesn't have a test for it

PR #1190, opened 2023-12-13
update templates

albho:


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,105 @@
...
+      options:
+        - label: I have read all of the relevant Picovoice Porcupine docs
+          required: true
+        - label: I have searched the existing issues for Porcupine
+          required: true
+  - type: dropdown
+    id: sdk
+    attributes:
+      label: SDK
+      options:
```
laves, line 24: Porcupine and Picovoice need `MCU` listed under SDKs as well. Microcontroller packages are only available for those two products I believe

laves (changes requested):

laves (approved):

PR #1361, opened 2025-02-04
Implement generation testing for Java

picoliu:


laves (changes requested):
We're also going to want to change the gradle version that the demo uses to check compatibility. See how this PR did it with gradle version: https://github.com/Picovoice/flutter-voice-processor/pull/20/files  (ignore the references to flutter version and agp version)

laves (changes requested):
sorry for the confusion - only the demo action should have the generation testing. The SDK and perf actions can stay just Java 11

laves (approved):

PR #1362, opened 2025-02-04
Update python minimum version

picoliu:


laves (approved):
this will also need a release, which we can do together

laves (approved):

PR #1363, opened 2025-02-04
Update react to 19.0

picoliu:


demo/react/package.json
```diff
@@ -7,14 +7,15 @@
...
-    "react-dom": "^18.2.0",
+    "@types/react": "^19.0.0",
+    "@types/react-dom": "^19.0.0",
+    "react": "^19.0.0",
+    "react-dom": "^19.0.0",
     "react-scripts": "5.0.1",
     "typescript": "^4.6.4"
   },
   "devDependencies": {
+    "@babel/plugin-proposal-private-property-in-object": "^7.21.11",
```
laves, line 18: why was this needed?

laves (commented):
did you also ensure all the warnings are gone and code is React 19 modern?

demo/react/package.json
picoliu: To eliminate this build warning:
```
One of your dependencies, babel-preset-react-app, is importing the
"@babel/plugin-proposal-private-property-in-object" package without
declaring it in its dependencies. This is currently working because
"@babel/plugin-proposal-private-property-in-object" is already in your
node_modules folder for unrelated reasons, but it may break at any time.

babel-preset-react-app is part of the create-react-app project, which
is not maintianed anymore. It is thus unlikely that this bug will
ever be fixed. Add "@babel/plugin-proposal-private-property-in-object" to
your devDependencies to work around this error. This will make this message
go away.
```

picoliu: I'll check React 19 changes and find out what can we apply to the demo.

laves: great, looks good then

PR #1367, opened 2025-02-05
Update demo package setup to read dependency versions from requirement file

picoliu:


laves (approved):
Can you release the demo package to pip as part of this? Same process as the SDK release

picoliu:
Done

PR #1425, opened 2025-06-09
rm respeaker demos

albho:


laves (approved):
You'll also need to:
- deprecate this package https://pypi.org/project/ppnrespeakerdemo/
- deprecate this package https://pypi.org/project/pvrespeakerdemo/
- find places we mention on the website/in the docs and delete or redirect

PR #1437, opened 2025-07-09
first toUpperCase explicit local setting in Android SDK

pico-mo:


pico-mo:
I messed up the branching for this I think, I'll start over

PR #1441, opened 2025-07-10
set locale explicitly to english in all toUpperCase calls under react-native

pico-mo:


ksyeo1010 (approved):

ksyeo1010 (commented):
you are missing package.lock files

.github/workflows/react-native-demos.yml
```diff
@@ -35,6 +35,10 @@ jobs:
       with:
         node-version: ${{ matrix.node-version }}
 
+    - name: Build pkg
```
ksyeo1010, line 38: revert

.github/workflows/react-native-demos.yml
```diff
@@ -60,6 +64,10 @@ jobs:
...
```
ksyeo1010, line 67: revert

PR #1469, opened 2025-10-16
React native v3.0.5

albho:
`Gradle 9+` no longer supports `jcenter()` and fails build for projects with dependencies that include `jcenter()`.

Updated `React Native` binding to not include `jcenter()`, and demo to be the latest `React Native` version, which uses `gradle 9.0`.


laves (commented):
Have you tested the demo on Android and iOS devices?

laves (approved):

PR #1503, opened 2025-12-11
v4.0

laves:


laves:
@tomlin7 - web services updated this morning. If you train a new keyword now, it will be compatible with v4.0 SDK versions. If you previously trained a keyword for v3.0, it will only be compatible with v3.0 SDK versions.

tomlin7:
@laves thanks for the response, I just retrained the model and I got the v4 ppn model file, thanks. 

Let me know if I could be of any help regarding documentation of the libraries/testing.

